PR #121, opened 2020-01-21
Feature/bug fix tp flash

EvenSol:


EvenSol:
A temporary fix proposed:
Use - NeqSim.jar   https://github.com/equinor/neqsim/releases/tag/v2.0.1

PR #125, opened 2020-01-29
New compressor chart type

marlenelund:
- Made interface for implementing different compressor chart classes
- Implemented CompressorChartAlternativeMapLookup which uses double interpolation with Fan laws to navigate the compressor map given flow and speed
- Implemented a constructor in the Compressor Class to support the changes
- Added a real test case (Åsgard A 27KA500) in both compressor chart classes
- Temporarily changed the Fan law correction code in CompressorChart class (created a bug for unequal number of points in different compressor curves in the map)


marlenelund:
@EvenSol Kan du hjelpe meg å finne ut av hvorfor Maven build feiler her?

PR #126, opened 2020-01-29
New compressor chart type

marlenelund:
- Made interface for implementing different compressor chart classes
- Implemented CompressorChartAlternativeMapLookup which uses double interpolation with Fan laws to navigate the compressor map given flow and speed
- Implemented a constructor in the Compressor Class to support the changes
- Added a real test case in both compressor chart classes
- Temporarily changed the Fan law correction code in CompressorChart class (created a bug for unequal number of points in different compressor curves in the map)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -172,6 +172,7 @@
         </dependency>
     </dependencies>
     <build>
+    <pluginManagement>
         <plugins>
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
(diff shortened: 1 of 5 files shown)


marlenelund:
NB! This interpolation method will return an error if the given flow is outside the flow range defined by the compressor map. That should probably be dealt with in some way..

EvenSol:
I have look at the code and it seems fine.
I did not understand this line:
double flowFanLaw = flow[i][j] * speed[i] / speed[0]; // MLLU: not correct. speed[0] should be the requested speed

PR #290, opened 2022-01-13
refact: typo atractive -> attractive

fannemel:
A lot of public classes and methods are changed so if this change is implemented, the version should be increased to notify there is no backwards compatability. An option might be to introduce copies and mark the ones with typos as deprecated first.


EvenSol:
I agree that we make a new version. I don't think backwards combability is a big problem at the moment (the backwards combability on serialized objects have been broken in earlier releases also).

EvenSol (approved):

PR #321, opened 2022-02-16
Feat allow serialization of process models

asmfstatoil:


EvenSol (approved):
Checked - and seems ok.

PR #323, opened 2022-02-17
Create ProcessSystemRunTransientTest.java

EvenSol:


EvenSol:
@asmfstatoil here is a test of dynamic calculation of a process. The tests fails now, but have worked at the time it was developed.

asmfstatoil:
included file in branch sim_class

PR #342, opened 2022-02-28
Rm more comp ctor

asmfstatoil:
Can't see that any of these constructors are in use


EvenSol (approved):
I think this ok - the constructors are not in use.

PR #355, opened 2022-03-07
refact: Standard classes: Create objects with name, avoid ctors without name

asmfstatoil:


EvenSol:
removed dispay() from test files

PR #373, opened 2022-03-26
Regression test property flash

asmfstatoil:


EvenSol (approved):
Great work!

EvenSol:
This is very good. The database has changed to H2, and some parameters in the model have been updated. So I guess this is the reason some test fails. We/I should go through the test to se it the failed test can be explained by the updated database parameters.

asmfstatoil:
Set in draft mode while troubleshooting failed tests.

EvenSol:
The expected results from this test can be updated (probably because of some minor update of parameters):
 ThermodynamicOperationsTest.testNeqSimPython2:79 array contents differ at index [5], expected: <4.910583053984447E-4> but was: <4.914277108631854E-4>

EvenSol (approved):

PR #389, opened 2022-04-13
controller initial work

EvenSol:


src/main/java/neqsim/processSimulation/processEquipment/ProcessEquipmentBaseClass.java
```diff
@@ -285,4 +286,18 @@ public boolean equals(Object obj) {
...
+     * @return the runTransient
+     */
+    public boolean isRunTransient() {
+      return runTransient;
+    }
+
+    /**
+     * @param runTransient the runTransient to set
+     */
+    public void setRunTransient(boolean runTransient) {
```
asmfstatoil, line 300: this should not be public

src/main/java/neqsim/processSimulation/processEquipment/ProcessEquipmentInterface.java
```diff
@@ -186,4 +186,11 @@
     /** {@inheritDoc} */
     @Override
     public int hashCode();
+    
+    public boolean isRunTransient();
+
+    /**
+     * @param runTransient the runTransient to set
+     */
+    public void setRunTransient(boolean runTransient);
```
asmfstatoil, line 195: should not be public

src/main/java/neqsim/processSimulation/processEquipment/ProcessEquipmentBaseClass.java
```diff
@@ -31,6 +31,7 @@ public abstract class ProcessEquipmentBaseClass extends SimulationBaseClass
     ControllerDeviceInterface flowValveController = null;
     public boolean hasController = false;
     private String specification = "TP";
+    private boolean runTransient = false;
```
asmfstatoil, line 34: should not be named the same as the function runTransient.

EvenSol: Good point. Will change that.

EvenSol:
if (getCalculateSteadyState()) 

if (getCalculateSteadyState()) is true - then the unit operation should be calculated as steady stat using normal run operation. For me it seems as the opposite is done (that transient is run if getCalculateSteadyState() is true.

asmfstatoil:
That is a bug/typo.

PR #395, opened 2022-04-28
set constant friction factor in pipelines simulation

EvenSol:


asmfstatoil:
todo: add more testing for setting and resetting wallcalcfrictionfactor.
todo: what about interphasefriction?
todo: verify that default methods don't have typos

PR #417, opened 2022-05-24
refact runTransient. Default to steady state calculation

asmfstatoil:


asmfstatoil:
@EvenSol Her har eg rydda i konseptet for å gjere steady state kalkulasjon når runTransient vert kalla.

EvenSol:
Looks like good clean up. Valves are initialized to be calculated dynamic (calcSteadyState=false), I guess this is ok. In the test they are set to be non-dynamic. Tee tests should by on valves running in dynamic mode I guess.

asmfstatoil:
the valves (and others with setcalulatesteadystate=false) are set up like that to keep their old behaviour if not specified. 

tests test the new behaviour

PR #426, opened 2022-05-30
refactor: Compressors have a single inlet and a single outlet and are twoports

asmfstatoil:
New interfaces and new abstract classes to improve organization of process simulation units


EvenSol (approved):
Dette ser fint ut!

Issue #444, opened 2022-06-14
Maven wrapper


MikAoJk:
**Is your feature request related to a problem? Please describe.**
 I'm always frustrated when a maven project do not use maven wrapper

**Describe the solution you'd like**
I would like this project to have a maven wrapper included 

**Describe alternatives you've considered**
Could rewrite the project to use gradle build tool

**Additional context**
On how to add inn maven wrapper from apache see: https://maven.apache.org/wrapper/

EvenSol:
Thanks for the input. Sounds like a good suggestion. Is this something you could help with @MikAoJk (add maven wrapper)?

MikAoJk:
Yes i can do that.
I will create a PR

PR #446, opened 2022-06-15
 Added in apache maven wrapper

MikAoJk:
#444


EvenSol:
Thanks a lot. The maven wrapper seem very nice! I did not use that before. I tested mvnw on linux, and it works great.

EvenSol (approved):

MikAoJk:
No problem. 
It is very nice to see that there is more and more source code open

PR #486, opened 2022-07-21
feat: add calcidentifier to simulation objects

asmfstatoil:
refact: moved property time up in hierarchy to be part of all simulation objects.


EvenSol (approved):
This seems fine.

PR #490, opened 2022-07-22
Log errors

asmfstatoil:


EvenSol (approved):
Great fixes👍

EvenSol:
I looked at the test that fail. It must be something with chemical reactions that have aviser it to fail I guess.

EvenSol:
Seems fine now👍

PR #533, opened 2022-09-30
refact: use useing pattern for disposable/close objects

asmfstatoil:


EvenSol (commented):
Great to clean up database operations!

PR #546, opened 2022-10-06
added serialization using Xstream

EvenSol:


EvenSol:
Xstream seems to be a good alternative to serialize neqsim objects. 
https://x-stream.github.io/index.html

asmfstatoil:
Ok, but where did we have issues with circular references now?

asmfstatoil (approved):
Serialization example works fine.

PR #557, opened 2022-10-22
556 set flow rate using unit ideal liquid sm3 as volume unit

EvenSol:


EvenSol:
Added unit: idSm3/day.

that can be used for method setTotalFLowRate(....)

This is the same as : Liq Vol Flow @ std conds in other process simulators.

PR #560, opened 2022-10-31
AddedSevereSlug

Sviatose:
AddedSevereSlug


Sviatose (commented):
Added Severe Slug

Sviatose:
Added the new tests. Now it references on the same stream object.

EvenSol (changes requested):
Very nice work :-)
Can you update so that it uses stream properties of input stream in calculations?

EvenSol (approved):

PR #585, opened 2022-11-24
chore: clean up workflows

asmfstatoil:


EvenSol (approved):
Seems like fine format update.

PR #615, opened 2023-04-09
feat: addedModuleSolutions

Sviatose:
This pull request adds a new class called ProcessModule to the NeqSim project. The class is designed to implement module calculations instead of running the whole process. The benefit of such implementation is that the user can define the modules in such a way that the recycling process does not overcalculate equipment. 

See the test for GlycolRig: /workspaces/neqsim/src/test/java/neqsim/processSimulation/processSystem/GlycolModulesTest.java

### DIFF
--- a/src/main/java/neqsim/processSimulation/processSystem/ProcessModule.java
+++ b/src/main/java/neqsim/processSimulation/processSystem/ProcessModule.java
@@ -1,0 +1,211 @@
+package neqsim.processSimulation.processSystem;
+
+import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


EvenSol:
Great! Did you compare the calculation time to a standard process?

Sviatose:
For the test I have provided: 
3 minutes and 54 seconds running as standard. 
2 minutes and 6 seconds running with modules.

EvenSol (approved):
Ready to be merged in
Master.

PR #616, opened 2023-04-09
fix: change logger.info to exception if no component is found in neqsim database

Sviatose:
Throw an exception if no component is found in neqsim database.

`throw new RuntimeException("No component with name: " + componentName + " in database");`


EvenSol (approved):
Seems good.

EvenSol:
Can be merged into master.

PR #620, opened 2023-04-14
hydrogen viscosity update

EvenSol:
viscosity of hydrogen returns NaN. Fix method.

Ref: https://github.com/equinor/neqsimhome/issues/3


EvenSol:
viscosity of hydrogen returns NaN. Fix method.

PR #651, opened 2023-05-01
updated maven pom

EvenSol:
updated pom for new release


eclipse_dictionary.txt
asmfstatoil, line 1: Not used if no one is using eclipse for developing.

asmfstatoil (approved):
It is fine to proceed with a new release now.

PR #658, opened 2023-05-03
refact: added enum phasetype

asmfstatoil:


EvenSol (approved):
I suggest we merge it into main.

PR #661, opened 2023-05-03
Phase type enum

asmfstatoil:


src/main/java/neqsim/physicalProperties/PhysicalPropertyHandler.java
```diff
@@ -101,20 +101,19 @@ public void setPhysicalProperties(PhaseInterface phase, int type) {
...
-      case "solid":
+      case SOLID:
         return solidPhysicalProperties;
-      case "wax":
+      case WAX:
         return solidPhysicalProperties;
-      case "hydrate":
+      case HYDRATE:
         return solidPhysicalProperties;
       default:
```
asmfstatoil, line 118: OBVIOUS that there are issues here, SOLIDCOMPLEX and LIQUID is missing

EvenSol (approved):
Good update👍

PR #670, opened 2023-05-09
chore: move main methods to test methods.

asmfstatoil:
@EvenSol The database tests are not working, should be looked at later.


EvenSol (approved):

EvenSol:
updated one test. The two others is ok to set Disabled.

PR #678, opened 2023-05-10
feat: added getUnit and copy methods to ProcessModule

Sviatose:
This pull request adds the `getUnit` and `copy` methods to the `ProcessModule` class. The `getUnit` method allows the user to retrieve a unit with a given name from the list of added unit operations and a list of added modules, while the copy method creates a deep copy of the `ProcessModule` instance.


EvenSol (approved):
Nice update👍

Issue #686, opened 2023-05-19
:bug: Bug - Three Phase Separator Producing Incorrect Output Phase


Sviatose:
**Bug Description**
I have encountered a bug related to the behavior of the three phase separator. The separator receives only the water phase but produce only oil phase as output.

`/workspaces/neqsim/src/test/java/neqsim/processSimulation/processSystem/waterDegasserTest.java`

 ```
@Test
  public void runProcess2() 
```

```
     neqsim.processSimulation.processEquipment.separator.ThreePhaseSeparator separator_VA_01 = new neqsim.processSimulation.processEquipment.separator.ThreePhaseSeparator(inlet_stream_VA_01);
     separator_VA_01.setName("FIRST_SEPARATOR");
     separator_VA_01.run();
 
     neqsim.processSimulation.processEquipment.heatExchanger.Heater TP_setter_hydrocyclone = new neqsim.processSimulation.processEquipment.heatExchanger.Heater("TP_SETTER_HYDROCYCLONE_AFTER_1ST_SEPARATOR",
     separator_VA_01.getWaterOutStream()
     );
     TP_setter_hydrocyclone.setOutPressure(61.0700675688386, "bara");
     TP_setter_hydrocyclone.run();
 
     neqsim.processSimulation.processEquipment.separator.ThreePhaseSeparator hydrocyclone_main = (new 
     neqsim.processSimulation.processEquipment.separator.ThreePhaseSeparator(TP_setter_hydrocyclone.getOutStream()));
     hydrocyclone_main.setName("HYDROCYCLONE_AFTER_THE_1ST_SEPARATOR");
     hydrocyclone_main.setEntrainment(0.01, "mole", "feed", "aqueous", "gas");
     hydrocyclone_main.run();
     // Bug here : received only water phase -> but produces only oil phase ?
     System.out.println(hydrocyclone_main.getLiquidOutStream().getFlowRate("kg/hr"));
     System.out.println(hydrocyclone_main.getWaterOutStream().getFlowRate("kg/hr"));
```

**To Reproduce**
1. Go to /workspaces/neqsim/src/test/java/neqsim/processSimulation/processSystem/waterDegasserTest.java 
2. Run test runProcess2

**Expected behavior**
System.out.println(hydrocyclone_main.getLiquidOutStream().getFlowRate("kg/hr")) should produce around 0, while System.out.println(hydrocyclone_main.getWaterOutStream().getFlowRate("kg/hr")) should produce around 33946

[...]

PR #692, opened 2023-05-21
Impovements and bug fixes for enum PhaseType

EvenSol:
Related to bug #686

Sume suggested updates for enum phase types.


EvenSol:
@asmfstatoil  Do we need to clone the enum PhaseType for a phase? How can we make sure the PhaseType follows a phase when it is cloned?

asmfstatoil:
> @asmfstatoil  Do we need to clone the enum PhaseType for a phase? How can we make sure the PhaseType follows a phase when it is cloned?

We should make a test to assert the required functionality. I thought enums were value types, can check tomorrow

EvenSol:
I think the bugs are fixed in the PR. Lets discuss if further improvements should be done.

PR #694, opened 2023-05-22
refact: clean up phasetype code

asmfstatoil:


src/main/java/neqsim/thermo/phase/Phase.java
```diff
@@ -80,7 +79,6 @@ public Phase clone() {
       logger.error("Cloning failed.", ex);
     }
 
-    clonedPhase.pt = PhaseType.byValue(pt.getValue());
```
EvenSol, line 83: are you sure we don't need to clone? I got the understanding that enums are passed by reference, and that we need to make a copy to make it independent from the object clone?

EvenSol (approved):
Good clean up. I need the understand why my auto formating is not the same as yours.

PR #696, opened 2023-05-22
170 improve method to set flow rate for a fluid

EvenSol:


EvenSol:
public void setTotalFlowRate(double flowRate, String flowunit) {
setTotalFlowRate(1.0)   - can be  solution where flow rate is low and init(1) gives an exception in volume calc
    init(0);
init(1);

PR #701, opened 2023-05-22
chore: mark deprecated constructors for removal

asmfstatoil:
refact: some constructors are optimized
style: some checkstyle violations are autofixed

Marking for removal means that the constructors shall be removed in the next major release, e.g., 2.6.0/3.0.0


EvenSol:
Making the code not compatible with Java 8 will mean that we can not use [IKVM](https://github.com/ikvmnet/ikvm) for compiling to .NET code. Else it is ok. But for the moment I think we should keep it compatible with Java 8.

PR #702, opened 2023-05-23
bugfix: left phase/component in halfway updated state when removing moles

asmfstatoil:


asmfstatoil:
Should have created issue first. Must review failing test for phProbe

asmfstatoil:
Fixed tests by allowing (and fixing) a mole reduction that would lead to a minimal negative number of moles in phase or for component in phase.

EvenSol (approved):
good fix

PR #703, opened 2023-05-23
refact: methods for adding moles to component and phase

asmfstatoil:


asmfstatoil:
The most interesting bit here is actually the documentation @EvenSol, please read the changes carefully to make sure the functions are described to do what is intended.

EvenSol (approved):

PR #706, opened 2023-05-24
refact: rm logfugacity as property

asmfstatoil:
docfix
sort method order to keep overloaded methods next to each other


EvenSol:
We need to check if this impacts performance. The reason for having logfuggacity coefficient was to limit the number of log calculations.

We should try to add methods to evaluate how changes affect performance (eg. the time taken for 1M flash calculations).

asmfstatoil:
Tried manually, this branch actually seems to be a bit faster 30792 vs 31903 ms  when running tpflash_benchmark with 50000 flashes.

EvenSol (approved):
it seems like it is ok.

EvenSol:
This can be merged into main I believe.

PR #707, opened 2023-05-24
refact: robustify h2 db init and updates from csv

asmfstatoil:
Had issues where testing failed and running main-methods failed due to not finding the csv-files.


EvenSol (approved):
Great to clean up the database use.

PR #716, opened 2023-05-29
715 implement check for need to recalculate unit operation

EvenSol:
Added check method for need for recalculation of unit operations (only added for streamSaturator and Stream up to now). This could be added to other unit operations as well. Seems like a good speed improvement is obtained for processes with recycles.


asmfstatoil (approved):
Would be nice with a test function

EvenSol:
> Would be nice with a test function

added some now

PR #717, opened 2023-05-30
Checkneedfor-recalc

EvenSol:


asmfstatoil:
What is the difference between solved (property of processequipmentbaseclass) and isupdated?

asmfstatoil:
There is probably a need for the functionality, I would just like to consider if the existing property for solved can be used.

EvenSol:
> There is probably a need for the functionality, I would just like to consider if the existing property for solved can be used instead of creating a new property and new getter.

Yes, that is propabaly a good idea. Will look into that.

EvenSol:
> There is probably a need for the functionality, I would just like to consider if the existing property for solved can be used instead of creating a new property and new getter.

updated to use solved

PR #718, opened 2023-05-30
test: add Chemicalreactionfactory test

asmfstatoil:
Also list the missing tables (some are no longer needed due to the moving of parameterfitting-code)


EvenSol:
The table reactionkspdata should be there.
https://github.com/equinor/neqsim/blob/master/src/main/resources/data/ReactionKSPdata.csv

I will need to check.

PR #732, opened 2023-06-08
chore: generate source and javadoc jar when building

asmfstatoil:


EvenSol (approved):
good to have doc and source included in build. Approved.

PR #738, opened 2023-06-15
fix: use correct iso standard for lng ship calculations

jorgenengelsen:
Seems like the logic to select which ISO standard version to use was inverted. Suggested a fix


EvenSol (approved):
Great bug fix! Thanks.

PR #741, opened 2023-06-16
refact: rewrote constructors for SystemThermo

asmfstatoil:
Removes a lot of discrepancies.


EvenSol (approved):
A lot of good fixes. The failed test can be solved independently.

asmfstatoil:
I dont really understand why it started failing now. 

The problem I see is that the reset call sets all component mole fractions to 0, which naturally creates issues as init(1) requires some component moles. Why it worked negler I dont know yet

PR #747, opened 2023-06-19
refact: reuse code from constructor in init++

asmfstatoil:


asmfstatoil:
init(0) calls reinitphasetype so also removed redundant calls

PR #752, opened 2023-06-23
refact: system getphase functions

asmfstatoil:


asmfstatoil:
This is still in draft state

EvenSol (changes requested):
we need to look into why some test faill.

EvenSol (approved):
👍

PR #753, opened 2023-06-24
Add missing database tables

EvenSol:
Added some of the missing database tables. Also updated the database connection for some of them


EvenSol:
It could be that we should have more NeqSimDatabase files. Havin only one, means that we would have to make h2 database from all csv files for all calculations. Eg. only doing a thermodynamic flash, material property tables used in mechanical design does not need to be added.

EvenSol:
> It could be that we should have more NeqSimDatabase files. By having only one, means that we would have to make h2 database from all csv files for all calculations. Eg. only doing a thermodynamic flash, material property tables used in mechanical design does not need to be added.

I have tried to improve this now by adding the design tables into its own database

PR #774, opened 2023-07-14
Pipe Beggs and Brills 

Sviatose:
Improvements:

Fixed the bug related to densities units.
Added unit tests to validate the code changes.
Added more output variables. 
Angle (or elevation or length are now calculated based on two other parameters)


EvenSol (approved):
Good update. Will be useful. Approved.

Sviatose:
We have discussed that pressureDrop method should return total pressure drop, thats I fixed in a last commit

PR #779, opened 2023-07-17
feat: added option to run isothermally pipe Beggs and Brills

Sviatose:
1. Made the variables private to ensure that their internal state is not directly accessible or modifiable by other classes.
2. Added option to run isothermally 
3. Added more get methods (angle, elevation)
4. fix: return total length instead of segment length


EvenSol:
I see Ine test of veggs and brille fail with a null printer exception. I can look into that later. Any idea why it happens ? @Sviatose

EvenSol (approved):

PR #782, opened 2023-07-25
Bug  dehydration glycol column test and bug fixes

EvenSol:


EvenSol:
I think the model now is ok. At low temperatures in the top, water are now able to get out of the column, and will flood the column?

PR #783, opened 2023-07-25
fix: Total length Pipe Beggs And Brills

Sviatose:
Encountered a problem when running simulation twice:

pipe.run()
pipe.run()

Fix total length calc. (used length from previous run). 
Also introduce the result matrix in run function for (so that the results will not get into the same lists).


EvenSol (approved):

EvenSol:
good update

PR #786, opened 2023-07-26
added manifold unit operation + tests

EvenSol:
A manifold is a process unit that can take in any number of streams and distribute them into a number of output streams. In NeqSim it is created as a combination of a mixer and a splitter.


Sviatose (approved):
It looks great! It's easy to understand

PR #788, opened 2023-07-26
update mechanical design simulations

EvenSol:
The automated mechanical design calculations needs to be updated. Tests need to be added, and design methods and data for more equipment need to be added.

The calculations should give a rough weight and volume estimation of a full process.


EvenSol:
@Sviatose to be discussed with you

PR #800, opened 2023-08-04
deactivate tests and improve test output

EvenSol:


EvenSol:
I have looked at the failed test. And I think the errors we see now can be explained/is reasonable. I support that we update the version.

PR #817, opened 2023-09-05
refactor: make abstract implementations of interfaces public

jorgenengelsen:
Needed to expose these abstract classes to generate typing stubs for NeqSimPython


EvenSol (approved):
This seems fine.

Issue #820, opened 2023-09-22
python neqsim calcproperties gives error in new package version


EvenSol:
**Describe the bug**
Fix bug : https://github.com/equinor/neqsimpython/issues/208

PR #821, opened 2023-09-22
fixed bug and added test

EvenSol:


EvenSol:
Ref:
https://github.com/equinor/neqsim/issues/820

PR #848, opened 2023-11-16
chore: remove non-implemented FileWriter classes

asmfstatoil:


EvenSol (approved):
I support that we remove these files.

PR #850, opened 2023-11-16
chore: remove redundant abstract class systemeos

asmfstatoil:


EvenSol:
I see that point that this class is not needed now, but it might be a correct structure for future. We use the same stutrcture in phase and component classes, so I think it is an advantage to keep this.

PR #868, opened 2023-12-03
updated dynamics calc

EvenSol:
Improve dynamics calculations and add tests


EvenSol:
@asmfstatoil 
Enig i begge kommentarer.

PR #877, opened 2023-12-07
added compressor test and initial code

EvenSol:
Add dynamic calculations for compressors

### DIFF
--- a/src/main/java/neqsim/processSimulation/processEquipment/compressor/Compressor.java
+++ b/src/main/java/neqsim/processSimulation/processEquipment/compressor/Compressor.java
@@ -1002,6 +1002,12 @@ public class Compressor extends TwoPortEquipment implements CompressorInterface 
     return getAntiSurge().isSurge();
   }
 
+  public double getDistanceToSurge() {
+    return (getInletStream().getFlowRate("m3/hr")
+        - getCompressorChart().getSurgeCurve().getSurgeFlow(getPolytropicFluidHead()))
+        / getCompressorChart().getSurgeCurve().getSurgeFlow(getPolytropicFluidHead());
+  }
+
   /**
    * <p>
    * isStoneWall.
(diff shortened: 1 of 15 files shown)


src/main/java/neqsim/processSimulation/processEquipment/compressor/Compressor.java
```diff
@@ -670,6 +670,44 @@ public void run(UUID id) {
...
+      inStream.getThermoSystem().init(3);
+      // inStream.run(id);
+    } catch (Exception e) {
+      logger.error(e.getMessage());
+    }
+    try {
+      outStream.getThermoSystem().setTotalFlowRate(actualFlowRateNew, "Am3/hr");
+      outStream.getThermoSystem().init(3);
+    } catch (Exception e) {
+      logger.error(e.getMessage());
```
jorgenengelsen, line 706: I noticed that this code segment uses a `try-catch` block to catch exceptions thrown by the methods of the ThermoSystem class.  It might be needed to silence exceptions in this case?

Just some suggestions:

1. Instead of catching all exceptions with catch (Exception e), consider catching only the specific exceptions that could be thrown by the setTotalFlowRate and init methods. This will make it easier to identify and handle errors that are specific to these methods.

2. Add validation tot the `actualFlowRateNew` to avoid causing exceptions. maybe it has to be not-null or positive?

3. Unless it is needed, it is good practice to throw the exceptions after logging it. This way it is less likely to cause issues for consumers of this code.

EvenSol: good suggestings. I think all are good ideas, and we should look into implementing this.

EvenSol: for the moment I work on getting the dynamic compressor to work, when it is working I will look into your suggestion.

EvenSol:
I have added a use case from Python bases in the tests:


https://github.com/EvenSol/NeqSim-Colab/blob/master/notebooks/process/dynamiccompressor.ipynb

src/main/java/neqsim/processSimulation/processEquipment/compressor/Compressor.java
EvenSol: @jorgenengelsen I have removed the try catch exception, and added a check for actualFlowRateNew.

jorgenengelsen: It looks good to me!

PR #886, opened 2023-12-20
added test for umrpru

EvenSol:
This PR add test to calculation of dew point pressure and dew point temperature and cricondenbar/cricondentherm conditions. It also fixes a bug in methods for direct calculation of dewpoint pressure/temperature that could make inaccurate calculations.

### DIFF
--- a/src/main/java/neqsim/PVTsimulation/simulation/SaturationPressure.java
+++ b/src/main/java/neqsim/PVTsimulation/simulation/SaturationPressure.java
@@ -31,7 +31,12 @@ public class SaturationPressure extends BasePVTsimulation {
    * @return a double
    */
   public double calcSaturationPressure() {
-    getThermoSystem().isImplementedCompositionDeriativesofFugacity(false);
+    boolean changed = false;;
+    if (!getThermoSystem().doMultiPhaseCheck()) {
+      changed = true;
+      getThermoSystem().setMultiPhaseCheck(true);
+    }
+    // getThermoSystem().isImplementedCompositionDeriativesofFugacity(false);
     getThermoSystem().setPressure(1.0);
     do {
       getThermoSystem().setPressure(getThermoSystem().getPressure() + 10.0);
@@ -62,6 +67,9 @@ public class SaturationPressure extends BasePVTsimulation {
     } while (Math.abs(maxPres - minPres) > 1e-5 && iteration < 500);
     getThermoSystem().setPressure(maxPres);
     thermoOps.TPflash();
+    if (changed) {
+      getThermoSystem().setMultiPhaseCheck(false);
+    }
     return getThermoSystem().getPressure();
   }
 
--- a/src/main/java/neqsim/PVTsimulation/simulation/SaturationPressure.java
+++ b/src/main/java/neqsim/PVTsimulation/simulation/SaturationPressure.java
@@ -31,9 +31,9 @@ public class SaturationPressure extends BasePVTsimulation {
    * @return a double
    */
   public double calcSaturationPressure() {
-    boolean changed = false;
+    boolean isMultiPhaseCheckChanged = false;
     if (!getThermoSystem().doMultiPhaseCheck()) {
-      changed = true;
+      isMultiPhaseCheckChanged = true;
       getThermoSystem().setMultiPhaseCheck(true);
     }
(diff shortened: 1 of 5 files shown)


src/main/java/neqsim/PVTsimulation/simulation/SaturationPressure.java
```diff
@@ -31,7 +31,12 @@ public SaturationPressure(SystemInterface tempSystem) {
    * @return a double
    */
   public double calcSaturationPressure() {
-    getThermoSystem().isImplementedCompositionDeriativesofFugacity(false);
+    boolean changed = false;;
```
jorgenengelsen, line 34: This ensures that multiphase check is always done for saturation pressure?

jorgenengelsen, line 34:
```suggestion
    boolean changed = false;
```

src/main/java/neqsim/PVTsimulation/simulation/SaturationTemperature.java
```diff
@@ -31,7 +31,11 @@ public SaturationTemperature(SystemInterface tempSystem) {
    * @return a double
    */
   public double calcSaturationTemperature() {
-    getThermoSystem().isImplementedCompositionDeriativesofFugacity(false);
+    boolean changed = false;;
```
jorgenengelsen, line 34:
```suggestion
    boolean isMultiPhaseCheckChanged = false;
```
Just a suggestion. Was not immediately clear to me, what "changed" at first

src/test/java/neqsim/PVTsimulation/simulation/SaturationTemperatureTest.java
```diff
@@ -1,51 +1,109 @@
...
+    tempSystem.init(0);
+    tempSystem.init(1);
+    // tempSystem.saveFluid(928);
 
-        SimulationInterface satPresSim = new SaturationTemperature(tempSystem);
-        satPresSim.run();
-        assertEquals(tempSystem.getTemperature(), 380.3071922, 0.1);
+    SimulationInterface satPresSim = new SaturationTemperature(tempSystem);
+    satPresSim.run();
+    assertEquals(tempSystem.getTemperature(), 380.3071922, 0.1);
```
jorgenengelsen, line 53:
```suggestion
    assertEquals(tempSystem.getTemperature(), 380.3, 0.1);
```
Since we now allow 380.3+-0.1, more decimals are not needed

src/test/java/neqsim/PVTsimulation/simulation/SaturationTemperatureTest.java
```diff
@@ -1,51 +1,109 @@
...
+    testSystem.setMixingRule("HV", "UNIFAC_UMRPRU");
+    testSystem.init(0);
+    ThermodynamicOperations testOps = new ThermodynamicOperations(testSystem);
+    try {
+      testOps.calcPTphaseEnvelope();
+    } catch (Exception ex) {
+      assertTrue(false);
+      throw new Exception(ex);
     }
+    assertEquals((testOps.get("cricondentherm")[0] - 273.15), 23.469396812206867, 0.02);
```
jorgenengelsen, line 101:
```suggestion
    assertEquals((testOps.get("cricondentherm")[0] - 273.15), 23.469, 0.02);
```

jorgenengelsen (approved):
Looks good. 

In practice this PR will make sure that saturation pressure and temperature will always be calculated with multiphase check?

Should not cause any large changes to results?

EvenSol:
> Looks good. Left some small comments and opinions.
> 
> In practice this PR will make sure that saturation pressure and temperature will always be calculated with multiphase check?
> 
> Should not cause any large changes to results?

Correct. It seems the flashes at high pressure are better when enabling multiphasecheck. SO for the moment it is best to force multiphasecheck in these calculations. But in principle to two phase flash should also work. I will look into that in the future. Added an Issue: #887

src/main/java/neqsim/PVTsimulation/simulation/SaturationPressure.java
EvenSol: yes. The multiphase flash is a more thorough flash (but slower). I have added an issue: #887

PR #898, opened 2024-01-08
fix: pipe beggs and brills y factor

Sviatose:
Fix factor S for multiphase flow when y value between 1 and 1.2


EvenSol (approved):
greit fix!

PR #902, opened 2024-01-12
fix: pipe BeggsAndBrills 

Sviatose:
- Change getSuperficialVelocity method to getInletSuperficialVelocity and getOutletSuperficialVelocity  
- Profiles start from segment 0 and length = 0 
- Add velocity to profile even if one phase 
- Add getNumberOfIncrements and getIncrementsProfile
- Change InvalidInputException to InvalidOutputException if calculated pressure is negative


EvenSol (approved):
Great update.

PR #911, opened 2024-01-29
build(deps): sync log4j version in all poms (#906)

EvenSol:
#906


asmfstatoil:
@EvenSol Maybe I misunderstood, is this branch actually for adding the test?

PR #914, opened 2024-02-08
fix: pipeBeggsAndBrills

Sviatose:
Solving the bug as discussed


Sviatose:
Still issues with PH flash. It produces me 2 phases while should be for clear only one. Looking into it

EvenSol (approved):
this seems good

PR #929, opened 2024-02-26
refact: use PhaseType enum

asmfstatoil:
do not use integer types to refer to phase type.


EvenSol (approved):
This is a great improvement for readability of code.

PR #934, opened 2024-02-27
refact: systemthermo.initAnalytic

asmfstatoil:


EvenSol (approved):
Great update! Both init method changes and initType make code better to read.

PR #936, opened 2024-02-27
test: add tests

asmfstatoil:


EvenSol (approved):
good test. All derivatives have not been implemented for PC-SAFT as you see.
Therefore properties requiring higher order derivatives might be wrong (eg. Cp), but I'm not sure. Fugacity and density should be ok. But I think how you have set it up is fine.

asmfstatoil:
All the test values are fine at my mac, but the values found in the build env are quite different so that speaks of some numerical issues. Especially seen for higher order derivatives.

Regarding the non-implemented functions, my intent here was to add tests before removing the methods that are purely redundant copies of code in parent class phasepcsaft.java. 

Interestingly I also saw that the phasepcsaftrahmat was in use for systempcsaft, not phasepcsaft

EvenSol:
> All the test values are fine at my mac, but the values found in the build env are quite different so that speaks of some numerical issues. Especially seen for higher order derivatives.
> 
> Regarding the non-implemented functions, my intent here was to add tests before removing the methods that are purely redundant copies of code in parent class phasepcsaft.java.
> 
> Interestingly I also saw that the phasepcsaftrahmat was in use for systempcsaft, not phasepcsaft

Yes - this is ok. THe model was developed as part of the PhD thesis by Rhamat W.:
https://ntnuopen.ntnu.no/ntnu-xmlui/handle/11250/2564750

asmfstatoil:
Nice, be free to merge it when you are ready

PR #941, opened 2024-03-07
refact: Phase classes clean up

asmfstatoil:


EvenSol (approved):
Seems like good updates😀

PR #962, opened 2024-03-19
feat: propertyFlash - more robust input handling 

asmfstatoil:


EvenSol (approved):
seems like a good update

PR #978, opened 2024-04-02
fix: update devcontainer.json to new format

asmfstatoil:
chore: update checkstyle version
style: fix xml and json tab settings


EvenSol (approved):
good update

PR #1037, opened 2024-06-28
fix: oil sep bug

Sviatose:
This should solve the bug with Oil Process


EvenSol (approved):
Seems like a great solution😀. It is noe test failing. Might be due to not being able to deserialize a class.

PR #1044, opened 2024-07-15
feat: update flow rate adjuster and fix negative change request if value is small (<1e-12)

Sviatose:
1. Update flow rate adjuster to accept 2 phases only. 
2. If the change request of a component dn is very small < 1e-12 and causes error (I got errors in splitter that component C12 1e-25 and change request was -2e-25 ..), set change request to 0 in this case.


EvenSol (commented):
looks fine

PR #1047, opened 2024-07-17
Cv method valves

EvenSol:


Sviatose:
Looks very nice ! Should we also add calculation in case of liquid / multiphase flow? The method you have added I assume is valid for gas volume fraction higher than 90%.

EvenSol:
> Looks very nice ! Should we also add calculation in case of liquid / multiphase flow? The method you have added I assume is valid for gas volume fraction higher than 90%.

@Sviatose I added the liquid Cv calcs

PR #1055, opened 2024-07-22
fix: Heat Exchangers wrong Enthalpy calculations

Sviatose:
When run HEX with UA value equal to 0, I got duty which was > 0 (and temperature change). This pull request fixes the problem by firstly running all the input streams


Sviatose:
See [bug--HEX-false-return-of-temperature](https://github.com/equinor/neqsim/tree/bug--HEX-false-return-of-temperature) for more info

PR #1084, opened 2024-08-22
fix: eclipse (read stuck at kij and aqueous phase)

Sviatose:


EvenSol (approved):
Great fix👍.

EvenSol (approved):
seems like a good solution

PR #1086, opened 2024-08-30
1078 implement gamma distribution model for plus fraction characterization

Sviatose:


EvenSol (changes requested):
@Sviatose maybe add in test an example of how to modify parameters of the gammamodel?

EvenSol (approved):

PR #1104, opened 2024-09-25
Named equipment test

asmfstatoil:


EvenSol (approved):
Good idea to force that a name is set on process equipment!

PR #1112, opened 2024-09-25
feat: new functions to clone and then set a new name

asmfstatoil:


asmfstatoil:
This will allow for an improved pattern for cloning and renaming streams in one go
`StreamInterface newStream = oldStream.clone(newName)`
implemented to avoid the following conversion to Stream object
`StreamInterface newStream = new Stream(newName,oldStream.clone)`

EvenSol (approved):
good suggestion

PR #1149, opened 2024-10-25
feat: Created physical property enum type to replace string

asmfstatoil:


src/main/java/neqsim/physicalproperties/PhysicalPropertyType.java
```diff
@@ -0,0 +1,34 @@
...
+import neqsim.util.exception.InvalidInputException;
+
+/**
+ * Types of PhysicalProperties, e.g. density, viscosity, conductivity.
+ *
+ * @author ASMF
+ */
+
+public enum PhysicalPropertyType {
+  DENSITY, VISCOSITY, CONDUCTIVITY;
```
EvenSol, line 12: DENSITY ➔ MASS_DENSITY
Reflects that it’s the density related to mass in a fluid context.

VISCOSITY ➔ DYNAMIC_VISCOSITY
Highlights that it’s the fluid’s resistance to flow.

CONDUCTIVITY ➔ THERMAL_CONDUCTIIVITY
Clarifies that it’s related to heat transfer rather than electrical conductivity.

EvenSol (commented):
suggest that we use more clear name for the physical properties:

DENSITY ➔ MASS_DENSITY
Reflects that it’s the density related to mass in a fluid context.

VISCOSITY ➔ DYNAMIC_VISCOSITY
Highlights that it’s the fluid’s resistance to flow.

CONDUCTIVITY ➔ THERMAL_CONDUCTIIVITY
Clarifies that it’s related to heat transfer rather than electrical conductivity.

THis has been a weakness in the current code also.

asmfstatoil:
I have renamed, but not updated the tests. Should it still be possible to write initPhysicalProperties("viscosity") with the old short names?

EvenSol:
> I have renamed, but not updated the tests. Should it still be possible to write initPhysicalProperties("viscosity") with the old short names?

Yes - It would be good to keep compatibility with old code, so maybe accept both viscosity and dynamic viscosity

PR #1161, opened 2024-11-03
change process package names

EvenSol:
Simplify package names in process package (avoid name lik processsimulation)


asmfstatoil (approved):

asmfstatoil:
Fint med enklare namn.

PR #1187, opened 2024-11-21
refact: use getComponent when getting 1 comp

asmfstatoil:


EvenSol (approved):
Yes support this. Could it be that computational speed is affected?

asmfstatoil:
> Yes support this. Could it be that computational speed is affected?

Not sure if Java JIT optimizer already handles this. But is probably faster when debugging.

EvenSol:
> > Yes support this. Could it be that computational speed is affected?
> 
> Not sure if Java JIT optimizer already handles this. But is probably faster when debugging.

It seems like the tests are running at the same speed - so propably this is just fine (and with no significant perfomrance change).

I suggest we merge this.

PR #1236, opened 2025-01-06
feat: new function phase.getComponentName

asmfstatoil:


EvenSol (approved):
Good update👍, guess it does not reduce speed?

PR #1250, opened 2025-01-10
refact: MixingRuleTypeInterface

asmfstatoil:
New Interface CPAMixingRuleType 
Introduce SuperInterface MixingRuleTypeInterface for EosMixingRuleType and CPAMixingRuleType 
Make mixingrule concept generic for phases, prepare for future cleaner split between types using eos mixing and cpa mixing


EvenSol (approved):
Seems like a good update!. I suggest you update the expected results in the test so they will pass (seem all of they are just minor changes).

asmfstatoil:
> Seems like a good update!. I suggest you update the expected results in the test so they will pass (seem all of they are just minor changes).

Found one more bug in my new code so after fixing that, there was no failing tests.

PR #1271, opened 2025-02-06
PFCT-bug-fix

akselrs:
PFCT viscosity model now works for all hydrogen mole fractions with methane.


EvenSol (commented):
Thanks lets have a short discussion on this. I think it seems good.

EvenSol (approved):

PR #1273, opened 2025-02-08
feat: update pipeBeggsAndBrills

Sviatose:


EvenSol (approved):
This looks fine.

Possible to turn on/off heat transfer calculations to improve speed when temperature changes is not important?

PR #1299, opened 2025-03-06
Compressor calculations with Leachman and Vega equation of state

victorigi99:
Added the possibility to do compressor calculations with Leachman and Vega

Added para and ortho-hydrogen as components and edited leachman accordingly


EvenSol (approved):
Good work👍

PR #1335, opened 2025-04-30
A couple of new models, and some chenges on LBC model for better accuracy for pure methane.

akselrs:
A couple of new models, and some chenges on LBC model for better accuracy for pure methane. The Muzny model is for pure hydrogen, and MuznyMod is more accurate.

### DIFF
--- a/src/main/java/neqsim/physicalproperties/methods/commonphasephysicalproperties/viscosity/LBCViscosityMethod.java
+++ b/src/main/java/neqsim/physicalproperties/methods/commonphasephysicalproperties/viscosity/LBCViscosityMethod.java
@@ -34,6 +34,9 @@ public class LBCViscosityMethod extends Viscosity {
   /** {@inheritDoc} */
   @Override
   public double calcViscosity() {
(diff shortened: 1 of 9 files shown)


src/main/java/neqsim/physicalproperties/methods/commonphasephysicalproperties/viscosity/LBCViscosityMethod.java
EvenSol, line 1: can the modifications be removed (because I understand it is all in a new file now?)

EvenSol (changes requested):
can the modifications in file: LBCViscosityMethod be removed?

src/main/java/neqsim/physicalproperties/methods/commonphasephysicalproperties/viscosity/LBCViscosityMethod.java
akselrs: Yes, everything is in a separate file, so you can completely ignore the LBCViscosityMethod.java file

PR #1341, opened 2025-05-09
add expander-compressor

EvenSol:
Turbo Expander-Compressor Simulation Plan

Given Inputs:

Inlet flow rate

Inlet pressure

Fluid composition

Inlet temperature

Target Output:

Desired outlet pressure after the expander

Use of equipment performance curves (e.g., efficiency and head)

Step-by-Step Calculation Logic:

1. IGV Modeling:

The Inlet Guide Vanes (IGV) are modeled as a valve with a Cv-like formula.

The pressure drop across the IGV depends on the opening position.

For example:
If inlet pressure is 65 bar and IGV opening is 80%, the outlet pressure after IGV might be approximately 63 bar.

2. Initial Speed Guess:

Start with an initial guess for the rotational speed (N, in rpm).

Calculate the following:

U/C ratio (tip speed ratio)

Q/N ratio (flow per unit speed)

Calculate the isentropic efficiency using the formula:
isentropic efficiency = design efficiency × correction factor 1 (based on U/C) × correction factor 2 (based on Q/N)

Use PS flash to calculate isentropic enthalpy at the outlet pressure.

Compute the isentropic work.

Calculate actual expander work:
expander work = isentropic efficiency × isentropic work

Use the enthalpy change to determine the outlet temperature and flow properties after the expander.

3. Mechanical Losses:

Calculate power losses due to bearings or other mechanical losses.

Example:
W_bearing = constant value (e.g., 1000 W), or based on speed.

4. Power Balance with Compressor:

Calculate the net power available to the compressor:
W_compressor_corrected = W_expander - W_bearing

Use performance curves to calculate:

Polytropic head correction

Polytropic efficiency

Required compressor power (W_compressor)

5. Iteration:

If W_compressor_corrected is not equal to W_compressor:

Adjust the rotational speed (N)

Repeat steps 2–4

Iterate until:
W_compressor_corrected ≈ W_compressor (within a given tolerance)

Convergence Criteria:

Stop iterating when the absolute difference in power is within a specified tolerance (e.g., 50 W)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -115,12 +115,7 @@
 			<artifactId>gson</artifactId>
 			<version>2.13.0</version>
 		</dependency>
-		<dependency>
-			<groupId>org.mockito</groupId>
-			<artifactId>mockito-core</artifactId>
-			<version>5.5.0</version>
-			<scope>test</scope>
-		</dependency>
+
 		<dependency>
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-databind</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -21,10 +21,10 @@
 	<licenses>
 		<license>
 			<name>The Apache Software License, Version 2.0</name>
-		  	<url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
-		  	<distribution>repo</distribution>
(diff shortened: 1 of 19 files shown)


EvenSol (commented):
update

PR #1357, opened 2025-06-04
add soreide method

EvenSol:


src/main/java/neqsim/thermodynamicoperations/ThermodynamicOperations.java
```diff
@@ -163,19 +163,104 @@ public void TPVflash(double volumeSpec, String unit) {
    * </p>
    */
   public void TPflash() {
+    // Check if system is Soreide-Whitson and use the special flash if so
+    if (system instanceof neqsim.thermo.system.SystemSoreideWhitson) {
```
asmfstatoil, line 167: instanceof is not available in java 8, must use some classname method

src/main/java/neqsim/thermo/phase/Phase.java
```diff
@@ -90,6 +90,11 @@ public abstract class Phase implements PhaseInterface {
   protected PhaseInterface[] refPhase = null;
   protected PhaseType pt = PhaseType.GAS;
 
+  /** Salinity of the phase (mol/kg or as used in Soreide-Whitson). */
```
asmfstatoil, line 93: Move to src/main/java/neqsim/thermo/phase/PhaseSoreideWhitson.java

EvenSol:
Can you do a thermodynamic cosistency check using:
https://github.com/equinor/neqsim/blob/master/src/main/java/neqsim/thermo/ThermodynamicModelTest.java

Sviatose:
> Can you do a thermodynamic cosistency check to see if the model is thermodynamic consitent using: https://github.com/equinor/neqsim/blob/master/src/main/java/neqsim/thermo/ThermodynamicModelTest.java
> 
> As was done here: https://github.com/equinor/neqsim/blob/b636c4d42f664b55653f1d4ce0b8b2de9cfe12e1/src/test/java/neqsim/thermo/system/SystemUMRPRUMCEosNewTest.java

PhaseEnvelope and TBPfraction failing

EvenSol:
@Sviatose 

I have tried to update to make test work.


What I think is strange:

  public static void setUp() {
    testSystem = new neqsim.thermo.system.SystemSoreideWhitson(298.0, 10.0);
    testSystem.addComponent("nitrogen", 0.7);
    testSystem.addComponent("ethane", 0.3);
    testSystem.init(0);
    testSystem.setMixingRule(11);
    
    why do we need to do    testSystem.init(0); before the mixing rule is set?

PR #1422, opened 2025-07-17
Feat: Gibbs reactor update

Sviatose:


EvenSol (approved):
Good uspdate.

what to look at later:
1. Use EJML for matrix computations (Jama is outdated, and we should use a more modern library when adding new models)
2. calculation of end temperature of Gibbs reactor assumes Cp is constant and heat of reaction and reaction equilibrium is calculated ar initial temperatures. Might be more accurate ways to du this.

Sviatose:
calculation of end temperature of Gibbs reactor assumes Cp is constant and heat of reaction and reaction equilibrium is calculated ar initial temperatures. Might be more accurate ways to do this. 

Temperature now is calculated every step of composition change. Cp is also calculated on everyTemperature change

Sviatose:
Test run with less then a second.

PR #1561, opened 2025-10-18
Separator level

EvenSol:
(https://github.com/equinor/neqsim/discussions/1559)

1- Included calculation for separator level from liquid volume(based on separator type) instead of the old percentage
2- Cross-section calculation for separator phases
3- Liquid volume calculation from actual liquid level
4- Getter and setter for liquid level updated to convert from percentage to original units.(both accept percentage)
5- Added new variables for radius, separator volume, and cross section
6- Updated class functions to implement those changes


EvenSol:
@copilot add tests for the updates

PR #1562, opened 2025-10-18
update separator level etc

EvenSol:
(https://github.com/equinor/neqsim/discussions/1559)

1- Included calculation for separator level from liquid volume(based on separator type) instead of the old percentage
2- Cross-section calculation for separator phases
3- Liquid volume calculation from actual liquid level
4- Getter and setter for liquid level updated to convert from percentage to original units.(both accept percentage)
5- Added new variables for radius, separator volume, and cross section
6- Updated class functions to implement those changes

### DIFF
--- a/src/main/java/neqsim/process/equipment/separator/Separator.java
+++ b/src/main/java/neqsim/process/equipment/separator/Separator.java
@@ -118,8 +118,8 @@ public class Separator extends ProcessEquipmentBaseClass implements SeparatorInt
   private double internalDiameter = 1.0;
   private double internalRadius = internalDiameter / 2;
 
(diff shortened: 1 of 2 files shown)


EvenSol:
@Alyasiri37 can you check the pul lrequest before i merge into master?

Alyasiri37:
@EvenSol Do you mean a final code review? If there is a reference process model from another tool then it might be safer to verify calculations before the merge.

EvenSol:
> @EvenSol Do you mean a final code review? If there is a reference process model from another tool then it might be safer to verify calculations before the merge.

no - it is just to see if your updates are still ok. The tests seems to run fine.

PR #1678, opened 2025-11-13
Add selectable step-response tuning rules with SIMC option 

EvenSol:
Replaced the step-response auto-tuning logic with Skogestad IMC (SIMC) formulations, providing PI and PID parameter calculations including robust handling of small dead times.

Added a configurable step-response tuning method with classic defaults and SIMC correlations inside ControllerDeviceBaseClass, including validation for invalid inputs. 

Exposed the StepResponseTuningMethod enum via the controller interface and broadened the auto-tuning test suite to cover classic, SIMC, and PI-only SIMC parameter sets.


EvenSol:
updated poms to solve issue with java8

PR #2098, opened 2026-04-21
Improvement on separator achitecture, where variables belong to the m…

dupuypablo:
Updated separators and scrubber architecture to having variables under mechanical design, and set-up methods delegated to scrubber/separator class.


EvenSol (approved):
good update

PR #2192, opened 2026-05-22
Update Distillation Column

Sviatose:
The solver first builds an initial column profile: feed flashes, pressure profile, estimated tray temperatures, and initial liquid/vapour flows.

It then solves the column equations: material balance, phase equilibrium, composition summation, and energy balance on all trays.

It iterates by correcting temperatures, flows, and compositions until residuals are small enough; if the simultaneous solver cannot produce an accepted result, the factory can fall back to a more robust damped substitution solver.

### DIFF
--- a/examples/notebooks/column_study.py
+++ b/examples/notebooks/column_study.py
@@ -1,36 +1,35 @@
 """Standalone column-study script.
 
 Mirrors the column_study.ipynb notebook but runs as a plain Python script:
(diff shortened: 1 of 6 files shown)


EvenSol (commented):
great! I tried to fix the test error. Can you check that it still is ok for you?

Sviatose:
Looks good to me

PR #2194, opened 2026-05-23
[codex] speed up flash calculations

EvenSol:
## Summary

This PR speeds up TP flash and reactive multiphase TP flash calculations with a small set of targeted changes:

- reuse the Armijo trial vector in `SysNewtonRhapsonTPflash`
- avoid repeating the enhanced multiphase stability check within one `TPmultiflash` run
- reuse iteration and linear-solve work arrays in `ModifiedRANDSolver`
- skip the expensive multiphase RAND solve for electrolyte cases where ionic species are only trace products after VLE initialization

## Why

The flash test suite was dominated by reactive multiphase electrolyte cases where the solver could spend thousands of iterations polishing trace ionic species. For these cases, the VLE phase split is already the material result and the total ionic feed is below `1e-8`, so the full multiphase RAND correction adds cost without changing the tested equilibrium behavior.

## Performance

Measured using Surefire XML suite time for the selected flash tests:

- Baseline: `188.813 s` across `225` flash tests, average `0.839 s/test`
- This branch: `74.322 s` across `225` flash tests, average `0.330 s/test`
- Improvement: about `60.6%` faster on the selected flash test suite

`ReactiveFlashBenchmarkTest` dropped from about `118.893 s` to about `1.257 s` in the full flash-suite run.

## Validation

Passed on the clean PR branch:

```text
.\mvnw.cmd test "-Dtest=neqsim.thermodynamicoperations.flashops.*Test,neqsim.thermodynamicoperations.flashops.saturationops.*Test,neqsim.thermodynamicoperations.flashops.reactiveflash.*Test" -DfailIfNoTests=false
```

Result:

```text
Tests run: 225, Failures: 0, Errors: 0, Skipped: 1
```

### DIFF
--- a/src/main/java/neqsim/thermodynamicoperations/flashops/SysNewtonRhapsonTPflash.java
+++ b/src/main/java/neqsim/thermodynamicoperations/flashops/SysNewtonRhapsonTPflash.java
@@ -60,6 +60,9 @@ public class SysNewtonRhapsonTPflash implements java.io.Serializable {
   /** u-variable array: u[i] = beta * y[i]. */
   private double[] uVector;
 
+  /** Trial u-vector used by the Armijo line search. */
+  private double[] uTrialVector;
+
   /** Pre-allocated EJML LU solver. */
   private transient LinearSolverDense<DMatrixRMaj> linearSolver;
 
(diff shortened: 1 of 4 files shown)


EvenSol:
@Sviatose skip the expensive multiphase RAND solve for electrolyte cases where ionic species are only trace products after VLE initialization

will this be aproblem?

PR #2261, opened 2026-06-07
Update Distillation Column

Sviatose:
Found some issues when integrating the column within the process object (process.run()). When doing recalculation streams inlet would double every time. Fixed.


EvenSol (approved):
Good fix👍

PR #2549, opened 2026-07-21
Consolidate equipment design behind typed kernels

EvenSol:
## Summary

- introduce `EquipmentDesignKernel<I, O>` as the standards-aware extension of the existing typed engineering calculation API
- add a registry that returns explicit `IMPLEMENTED` / `NOT_IMPLEMENTED` status instead of implying support
- adapt the legacy mutable API 610 calculator through a pure kernel that defensively copies its configuration
- return an immutable API 610 assessment snapshot with calculation provenance and review-gated status
- expose kernel lookup through `StandardRegistry` while preserving all existing `MechanicalDesign` and calculator entry points
- make API 610 edition support fail closed and align the catalog with the implemented 13th-edition screening basis

## Engineering boundary

The API 610 implementation remains `SCREENING`. It does not certify conformity and always requires purchased-standard, project, engineering, and vendor review. The kernel currently accepts only 13th edition; other editions are rejected as `EDITION_NOT_IMPLEMENTED`.

## Compatibility

The adapter does not mutate a caller-owned calculator or `ProcessSystem`. Legacy API 610 usage through `PumpMechanicalDesign` is unchanged. No API 610 numerical criteria or equations are changed in this PR.

## Stacking

This draft is intentionally based on `agent/typed-design-standard-selection` (PR #2547), which is based on PR #2542. It should be rebased and retargeted as the earlier stages merge.

## Validation

- changed core sources compile with the JDK compiler harness
- smoke test executes a complete passing API 610 assessment and verifies the legacy input remains unmodified
- focused test source compiles against the changed core API
- generated standards matrix matches the published documentation exactly
- GitHub Actions provides the authoritative Spotless and repository-level validation
## CI baseline note

- `Spotless format patch`, `PaperLab Replication Gate`, and `Execute complete offshore engineering notebook` pass on the final head.
- The process-to-engineering notebook workflow compiles all production and test sources past this branch's changes, then stops at the unchanged `master` method reference in `FreezingPointTemperatureFlashTest`, where JUnit reports an ambiguous `assertDoesNotThrow` overload.
- The branch-caused null-overload ambiguity in `StandardRegistryTest` was fixed on both stacked branches.

PR #2547, opened 2026-07-21
Add typed fail-closed design-standard selection

EvenSol:
## Summary

- add immutable `StandardEdition` and `StandardSelection` APIs for explicit editions and project amendments
- add structured `StandardApplicability` results and machine-readable `StandardSelectionException` reasons
- make strict registry selection fail closed for catalog-only, disconnected, missing-context, and inapplicable standards
- preserve the existing permissive `StandardType` factory overloads and provide an explicit legacy-compatible migration mode
- record whether audited calculations are actually reachable through `StandardRegistry`
- document the strict/legacy boundary and add focused regression coverage

## Compatibility

This stage is additive. Existing `StandardRegistry.createStandard(StandardType, ...)` and `MechanicalDesign.setDesignStandard(StandardType, ...)` behavior is unchanged. No numerical design equations are changed.

## Stacking

This draft is intentionally based on `agent/audit-design-standard-support` (PR #2542). After that PR merges, this branch can be rebased and the PR retargeted to `master`.

## Validation

- focused tests cover immutable edition data, amendments, explicit-edition reproducibility, structured applicability, each strict rejection class, mechanical-design integration, and legacy-compatible fallback
- local Maven execution is unavailable in the sandbox because dependencies cannot be downloaded; full build, tests, Javadoc, Spotless, and pre-commit validation are delegated to GitHub Actions

PR #2542, opened 2026-07-20
Audit mechanical-design standard support

EvenSol:
## Summary

- add `StandardSupportLevel`, `StandardSupport`, and `StandardSupportAudit` to distinguish catalog metadata from executable engineering calculations
- expose the class selected by `StandardRegistry` without constructing equipment or reading the design database
- classify all 32 current `StandardType` entries: 25 preliminary screening entries and 7 catalog-only entries; none are presented as validated or qualified
- publish a generated support matrix covering edition metadata, category, registry mapping, calculation path, maturity, and the current engineering boundary
- add regression coverage that requires every standard to be represented and keeps the published matrix identical to the generated source

## Why

The existing catalog and guide used “supported” for both standard metadata and numerical implementations. Several catalog entries map only to the base `DesignStandard`, while others map to generic category calculators that do not implement an edition-specific conformity assessment. The previous guide also contained stale or nonexistent API examples.

This change makes the current evidence explicit before later PRs introduce typed editions, fail-closed selection, and qualified equipment kernels.

## Important findings made explicit

- API 610 screening exists through `PumpApi610DesignCalculator`, but `StandardRegistry` currently selects only the base `DesignStandard`.
- API 650/620 are catalogued but map to a separator-oriented pressure-vessel class; no tank-code calculation is claimed.
- API 660/661 and ISO 16812 have no connected standard-specific exchanger calculation.
- API 521/526 map to `ValveDesignStandard`, which does not implement the named relief calculations.
- existing vessel, piping, separator, compressor, and material paths are identified as preliminary screening with their limitations.

## Compatibility and numerical behavior

- the existing `createStandard(...)` behavior and returned classes are preserved
- no equipment sizing equation, default, serialization field, or process-simulation result is changed
- unsupported selections do not fail yet; that behavior belongs in the follow-up typed-standards PR
- the matrix describes NeqSim implementation evidence and does not reproduce standards content or claim certification

## Validation

Completed locally:

- Java 8 production-source compilation for the changed registry and audit model using the JDK compiler module and isolated dependency stubs
- Java 8 test-source compilation for `StandardSupportAuditTest`
- executable generation of the 32-row Markdown matrix
- byte-for-byte comparison of the generated matrix with the published documentation block
- verified the existing documentation index already references the updated page and the new internal engineering link exists

Hosted evidence:

- the repository Spotless workflow produced an exact formatter patch, which was applied mechanically to all reported files
- documentation search and PaperLab passed on the initial head
[...]

PR #2551, opened 2026-07-21
Preserve system mechanical design state

EvenSol:
## Summary

- preserve configured equipment `MechanicalDesign` instances during system-wide initialization, execution, and equipment-specific access
- persist the cached `SystemMechanicalDesign` with `ProcessSystem` copies and Java serialization
- expose whether aggregation has run and a monotonic calculation revision
- return deep defensive equipment-summary snapshots and retain numeric dimensions
- correct total plot space from module height × length to module width × length
- remove direct console output from mechanical-weight aggregation

## Correctness boundary

This PR intentionally preserves the existing per-equipment exception handling in `SystemMechanicalDesign`. Structured failure propagation is the next correctness slice. It does not change any equipment sizing equations or standards implementation.

## Compatibility

Existing public entry points remain available. `initAllMechanicalDesigns()` is now idempotent for already initialized equipment, so caller-configured standards, limits, and sizing inputs are retained. Older serialized `ProcessSystem` objects remain compatible because the added persisted aggregate field defaults to `null` and is lazily recreated.

## Stacking

This draft is intentionally based on `agent/design-kernel-api-consolidation` (PR #2549), which is based on PR #2547 and PR #2542. It should be rebased and retargeted as the earlier stages merge.

## Validation

- Java 8 compiler harness compiles the changed `SystemMechanicalDesign` against minimal dependency stubs
- behavioral smoke test covers repeat execution, configured-object identity, stable totals, corrected footprint, defensive summaries, Java serialization, and restored process back-reference
- repository parse reached dependency resolution without syntax errors; the local environment cannot download Maven dependencies
- GitHub Actions is the authoritative Spotless and repository-level validation

### DIFF
--- a/docs/process/mechanical_design.md
+++ b/docs/process/mechanical_design.md
@@ -161,6 +161,10 @@ sysMecDesign.setCompanySpecificDesignStandards("Equinor");
 // Run design calculations for all equipment
 sysMecDesign.runDesignCalculation();
 
+// The aggregate is cached on the ProcessSystem and survives copy/serialization
+boolean calculated = sysMecDesign.hasRunDesignCalculation();
+long revision = sysMecDesign.getDesignCalculationRevision();
+
 // Access aggregated results
 double totalWeight = sysMecDesign.getTotalWeight();           // kg
 double totalVolume = sysMecDesign.getTotalVolume();           // m³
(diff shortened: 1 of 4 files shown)


EvenSol:
Included in the consolidated process-design update #2561, which passed all hosted validation workflows and was squash-merged into `master` as `47a1aaaea72dcfb92c5ef946c68ea25b19cba151`. Closing this stacked slice as superseded; its final content is present on `master`.

PR #2623, opened 2026-07-26
fix: correct vessel validation pressure units

EvenSol:
## Summary

- treat `SystemInterface.getPressure()` as bar in `VesselDepressurization` validation
- prevent false backpressure errors for ordinary high-pressure vessels
- restore the intended warning for pressures above 700 bar
- return the common non-critical configuration warnings from `validateWithWarnings()`
- add regression coverage for pressure units, warning propagation, and error/warning separation

## Defect

`thermoSystem.getPressure()` already returns bar, but the validation paths divided it by `1e5` as though it were pascal. An 80 bar vessel was therefore interpreted as 0.0008 bar, so a valid 1.2 bar backpressure failed validation. The same conversion prevented the high-pressure equation-of-state warning from firing.

The existing `testValidateWithWarningsMethod` passed only because that false pressure error was returned as an `ERROR:` entry. Once the pressure conversion was corrected, the test exposed a second defect: `validate()` collected and logged common configuration warnings, but `validateWithWarnings()` discarded them. The shared `collectConfigurationWarnings()` path now keeps both validation entry points consistent.

## Validation

- Windows Java 21 `VesselDepressurizationTest`: 49 tests, 0 failures, 0 errors, 0 skipped
- Windows Java 21 complete suite: `BUILD SUCCESS` in 43:48
- pre-commit: passed
- Spotless: passed
- CodeQL: passed
- PaperLab replication gate: passed
- Javadoc and agent-skill reference checks: passed
- Ubuntu Java 21 JaCoCo suite: still running on head `88bc1cb3`


EvenSol:
Copilot review assessed on final implementation:

- Accepted the inline test-hardening suggestion in `7433f1c`: the 750-bar case now rejects any `ERROR:` result before asserting the high-pressure warning.
- Accepted both suppressed unit-naming suggestions: validation locals are now `initialPressureBar`, making the `SystemInterface`/backpressure unit boundary explicit.
- The original thread is resolved; the complete two-file diff will be revalidated by Spotless, pre-commit, the targeted/full test build, CodeQL, and final-head Copilot review.

PR #2700, opened 2026-07-30
fix: run each controller once per transient step

EvenSol:
## Root cause

`ProcessSystem.runTransient(dt, id)` executes equipment first and then scans controllers registered through `ProcessSystem.add(controller)`. A controller integrated by equipment and also registered for discovery was therefore advanced twice with the same physical timestep. Repeated standalone registration had the same effect. Under `SEMI_IMPLICIT`, equipment is deliberately evaluated twice with one calculation identifier, which could advance a built-in PID yet again.

The first patch attempted to reserve every controller attached to equipment. Review correctly identified that attachment does not prove execution: equipment such as `Heater` exposes controller association but does not integrate it in `runTransient`. Reserving attachments could suppress a controller entirely.

## Scope

- add the backward-compatible default `ControllerDeviceInterface.hasRunTransient(UUID)` contract
- make `ControllerDeviceBaseClass.runTransient(..., id)` idempotent for one non-null calculation identifier
- coalesce the standalone scan by object identity
- skip the standalone call only when the controller reports that exact timestep identifier
- preserve standalone execution for controllers merely attached to non-integrating equipment
- preserve distinct same-name controller objects and repeated execution on later timesteps
- document the contract in Javadocs, the dynamic guide, and the dynamic-simulation skill

No method is removed or behavior changed for custom controller implementations that do not opt into identifier tracking; the new interface method defaults to `false`.

## Baseline and before/after evidence

A deterministic counting-controller harness uses methane process cases and calculation identifiers, avoiding wall-clock assertions.

| Case | Current `master` | This branch |
|---|---:|---:|
| Dynamic valve controller also registered standalone, explicit step | 2 updates/timestep | 1 |
| Same valve case, semi-implicit equipment pass | 3 updates/timestep | 1 |
| Same standalone identity registered twice | 2 updates/timestep | 1 |
| Controller attached to a heater that does not execute it | 1 standalone update | 1 |

The heater case is the regression for the review finding: association alone must not be treated as execution.

## Engineering behavior

Controller state ownership is determined by both object identity and the calculation UUID:

- identity coalesces duplicate entries in the standalone registration list;
- the UUID proves that equipment actually completed the controller update in this timestep;
- `ControllerDeviceBaseClass` rejects repeated calls with that UUID, including semi-implicit re-evaluation;
- a later timestep uses a different UUID and advances normally;
- custom controllers remain conservative by default and can implement the same contract when equipment integrates them.

Controller timing relative to equipment, actuator coupling, registration/discovery lists, serialization fields, event scheduling, measurements, history, process time, thermodynamic calculations, initialization levels, tolerances, mass/energy equations, and phase behavior are otherwise unchanged.

## Tests and validation

Focused regression class: `ProcessSystemTransientControllerTest` (5 tests)

- equipment-executed plus standalone registration
- semi-implicit repeated equipment pass
- attachment-only heater fallback
- duplicate standalone identity
- distinct same-name identities across repeated timesteps

On repaired head `ade5eaf5`, every hosted check passes:

- focused `ProcessSystemTransientControllerTest` coverage is included in the successful fast-test suites
- Java 21 fast tests pass on Ubuntu and Windows
- Java 8 tests pass on Ubuntu and Windows
- all three Java 21 slow-test shards pass
- Javadocs, Spotless, pre-commit, skills/agents lint, PaperLab, documentation search, and CodeQL pass

No check was rerun selectively or weakened, and no tolerance or convergence criterion was changed.

## Compatibility and risk

- additive Java API only, implemented as a default method
- existing third-party controllers continue to execute conservatively unless they opt into identifier tracking
- built-in controller repeated calls with the same non-null UUID are now idempotent
- null identifiers are not coalesced
- distinct controller instances remain independent even when names match
- inactive controller behavior is unchanged
- no test tolerance or convergence criterion is changed

## Documentation impact

Updated:

- `ControllerDeviceInterface` and `ControllerDeviceBaseClass` Javadocs
- `ProcessSystem.add(ControllerDeviceInterface)` Javadoc
- `docs/process/dynamic-simulation-enhancements.md`
- `.github/skills/neqsim-dynamic-simulation/SKILL.md`

## Related work

Builds on merged transient execution fixes #2686, #2692, and #2695. No other open dynamic-simulation PR overlaps this change.

## Next dependency

Define explicit transient convergence semantics for recycle and implicit signal couplings, followed by incomplete-step rollback/state ownership.

### DIFF
--- a/.github/skills/neqsim-dynamic-simulation/SKILL.md
+++ b/.github/skills/neqsim-dynamic-simulation/SKILL.md
@@ -34,11 +34,13 @@ Each timestep:
    order or dependency-aware graph levels when parallel transient execution is
    enabled. Setter units are excluded from these explicit, semi-implicit, and
    parallel equipment passes.
-4. Standalone controllers run; equipment-embedded controllers retain their
-   equipment-specific execution timing for compatibility. Controller ownership
-   is identity-based: registering an embedded controller separately, or adding
-   one standalone controller instance repeatedly, does not update its PID state
-   more than once per timestep.
+4. Standalone controllers run; controllers actually executed inside equipment
+   retain their equipment-specific timing for compatibility. Execution is
+   coalesced by object identity and the timestep calculation identifier. Merely
+   attaching a controller to equipment that does not execute it must not suppress
+   the standalone update; repeated standalone registration runs once per step.
+   Custom controllers integrated by equipment must report the completed identifier
+   through `hasRunTransient(UUID)`.
 5. Measurement devices are sampled, alarms are evaluated, and history is stored.
 
 ## Basic Dynamic Setup
--- a/.github/skills/neqsim-dynamic-simulation/SKILL.md
+++ b/.github/skills/neqsim-dynamic-simulation/SKILL.md
@@ -35,7 +35,10 @@ Each timestep:
    enabled. Setter units are excluded from these explicit, semi-implicit, and
    parallel equipment passes.
 4. Standalone controllers run; equipment-embedded controllers retain their
-   equipment-specific execution timing for compatibility.
+   equipment-specific execution timing for compatibility. Controller ownership
+   is identity-based: registering an embedded controller separately, or adding
+   one standalone controller instance repeatedly, does not update its PID state
+   more than once per timestep.
 5. Measurement devices are sampled, alarms are evaluated, and history is stored.
 
 ## Basic Dynamic Setup
--- a/.github/skills/neqsim-dynamic-simulation/SKILL.md
+++ b/.github/skills/neqsim-dynamic-simulation/SKILL.md
@@ -39,8 +39,10 @@ Each timestep:
    coalesced by object identity and the timestep calculation identifier. Merely
    attaching a controller to equipment that does not execute it must not suppress
    the standalone update; repeated standalone registration runs once per step.
-   Custom controllers integrated by equipment must report the completed identifier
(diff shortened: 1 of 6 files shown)


EvenSol (commented):
Copilot feedback was technically valid and has been addressed in the latest head. The implementation and documentation now distinguish controllers actually executed inside equipment from controllers merely attached for association. Attachment-only controllers still run in the standalone phase; equipment-integrated controllers coalesce via the timestep calculation identifier. Focused regression coverage includes both a valve-owned controller and a Heater attachment-only fallback. All three Copilot threads are resolved.

EvenSol (commented):
Final feedback disposition for head `ade5eaf5`:

- The three original attachment-versus-execution findings were valid, fixed, regression-tested, and their threads are resolved.
- Copilot's later full review covered all six changed files and produced no new actionable comments.
- The suppressed mixing-rule suggestions do not apply to these pure-methane controller-counting tests: no binary interaction or mixture rule enters the asserted behavior, and the cases passed all Java 8/21 Ubuntu/Windows matrices.
- The suppressed two-argument overload note is intentional API behavior: `runTransient(response, dt)` creates a new identifier and therefore denotes an independent update. Coalescing requires propagation of the same process timestep UUID through the three-argument overload, as documented by the new contract.

No further code change is justified by the remaining low-confidence suggestions.

Issue #2705, opened 2026-07-31
TwoFluidPipe transient closure relaxation can overwrite conserved phase mass


EvenSol:
## Problem

On current `master` (c80a31794ac2edc71d697324e966d38b8eec5bc4), `TwoFluidPipe.runTransient(...)` first advances the seven conservative variables, but then calls `relaxHoldupTowardSteadyClosure(dt)`. That method changes primitive gas/liquid holdups and velocities and finally calls `TwoFluidSection.updateConservativeVariables()`, which reconstructs phase masses from the changed primitive state:

```java
gasMassPerLength = alphaG * rhoG * A;
liquidMassPerLength = alphaL * rhoL * A;
```

The projection is not represented as a conservative face flux or an equal-and-opposite phase-transfer source. It can therefore create or remove total pipe inventory after the conservative time step.

Relevant paths:

- `src/main/java/neqsim/process/equipment/pipeline/TwoFluidPipe.java`
  - end of `runTransient(...)`
  - `relaxHoldupTowardSteadyClosure(...)`
- `src/main/java/neqsim/process/equipment/pipeline/twophasepipe/TwoFluidSection.java`
  - `updateConservativeVariables()`

No open issue or current open PR was found for holdup-relaxation conservation, transient inventory conservation, or this method name. PR #2697 is already merged and concerns three-layer geometry, not transient conservation.

## Exact analytical baseline

For a two-phase cell with fixed area and densities, changing liquid holdup from \(\alpha_L\) to \(\alpha_L'\), with \(\alpha_G=1-\alpha_L\), changes reconstructed mixture mass per unit length by

$$\Delta m' = A(\rho_L-\rho_G)(\alpha_L'-\alpha_L).$$

The current relaxation uses

$$\alpha_L'=\alpha_L+\left(1-e^{-\Delta t/4\,\mathrm{s}}\right)(\alpha_{L,\mathrm{closure}}-\alpha_L).$$

A deterministic illustrative cell using the implementation equations:

- diameter: 0.2 m, so \(A=0.0314159265\,\mathrm{m^2}\)
- \(\rho_G=40\,\mathrm{kg/m^3}\)
- \(\rho_L=700\,\mathrm{kg/m^3}\)
- initial \(\alpha_L=0.05\)
- closure target \(\alpha_{L,\mathrm{closure}}=0.15\)
- requested step: 1 s

gives \(\alpha_L'=0.0721199\) and

[...]

PR #2718, opened 2026-07-31
fix: preserve TwoFluidPipe mass after transient closure update

EvenSol:
## Summary

Fixes the foundational transient state-ownership defect in #2705 by removing two post-step paths that rebuilt finite-volume mass from primitive holdup:

1. the unreferenced 4 s relaxation toward steady-state holdup; and
2. the unconditional primitive-to-conservative synchronization after slug/terrain tracking.

Gas, oil, and water masses advanced by the finite-volume equations remain authoritative. The PR also adds `getTotalMassInventory()` for direct balance diagnostics.

## Reproducible evidence

The removed closure projection used:

```text
alphaL* = alphaL + [1 - exp(-dt / 4 s)] (alphaL,steady - alphaL)
mL'     = alphaL* rhoL A
mG'     = (1 - alphaL*) rhoG A
```

without a boundary flux or mass-transfer source. For D = 0.20 m, rhoG = 40 kg/m3, rhoL = 700 kg/m3, alphaL = 0.05, target alphaL = 0.15, and dt = 1 s, it changes total mass by 0.4586 kg/m, or about 459 kg over 1 km.

A hosted 3 km regression initialized at 6 kg/s and stepped to 12 kg/s for 1 ms then exposed the second path: inventory changed by 1925.252 kg although the conservative inlet-plus-outlet flux envelope was only 0.036 kg. The default Lagrangian pass made no accepted conservative transfer, but `synchronizeConservativeStateWithPrimitiveState()` still called `updateConservativeVariables()` on every section and recreated domain-volume-scaled mass.

## Governing balance

```text
M(t+dt) - M(t) = integral_t^(t+dt) [mDot_in - mDot_out] dt
M = sum_i (mGas'_i + mOil'_i + mWater'_i) dx_i
```

Phase mass per length is kg/m, section length is m, inventory is kg, boundary flow is kg/s, and time is s. Internal phase transfer must be equal and opposite and cancel in total mass.

## Implementation

- Remove `relaxHoldupTowardSteadyClosure()` and its post-step pressure/state reconstruction.
- Remove unconditional slug/terrain primitive-to-conservative synchronization.
- Keep PDE flux integration, thermodynamics, boundary APIs, steady-state initialization, and public signatures unchanged.
- Add read-only `getTotalMassInventory()`.
- Preserve deterministic calculation identity in the regression.

Auxiliary slug and terrain trackers can retain diagnostic primitive state, but cannot promote it into conservative phase mass without an explicit conservative flux or source formulation.

## Pressure versus inventory settling

After the conservative correction, the mass regression and one-phase pressure cases pass. The two- and three-phase pressure disturbances also reduce from 600,000 Pa to 9,757 Pa and 16,444 Pa respectively, both far inside the unchanged 200,000 Pa pressure limit.

The remaining failures came only from also requiring liquid-holdup RMS <= 0.08 inside the same 60 s acoustic-pressure test. That is not a valid shared timescale: pressure waves settle acoustically, while liquid inventory approaches a stationary distribution by material transport and may require one or more residence times. The regression now gates the 60 s scenario only on the unchanged pressure criterion; holdup RMS remains calculated and reported diagnostically. The strict mass-flux envelope is unchanged.

## Validation

Current head: `816eb5ebde2520ba17dafa9ef832ba22683afabb`.

Coverage includes:

- 6 -> 12 and 6 -> 30 kg/s rate steps over 1 ms;
- one-, two-, and three-phase outlet-pressure steps;
- independent stationary pressure targets;
- pressure, liquid-, oil-, and water-holdup diagnostics;
- full repository Java 8/21, Ubuntu/Windows, and slow-test matrices.

On the exact current head, Java 21 Ubuntu and Windows, all three Java 21 slow-test shards, Javadocs, Spotless, pre-commit, CodeQL, PaperLab, documentation search, and agent-skill checks pass. The Java 8 Ubuntu and Windows jobs are still running. The immediately preceding head passed both Java 8 jobs; the only later source change corrected a diagnostic test log from "reached new stationary state" to "pressure profile settled." All Copilot findings have been dispositioned and no review threads are open.

## Compatibility and risk

Public APIs remain compatible. The production change removes an unbalanced state projection; it does not change correlations, tolerances, phase equilibrium, or boundary values. It does not claim experimental validation of pressure, holdup, or slug frequency.

## Documentation impact

Updated `TwoFluidPipe` Javadocs and `docs/process/TWOFLUIDPIPE_MODEL.md` with the inventory equation, units, diagnostic API, state ownership, and the distinction between acoustic pressure settling and material-inventory settling.

## References

- Bendiksen et al. (1991), *The Dynamic Two-Fluid Model OLGA: Theory and Application*: https://onepetro.org/PO/article/6/02/171/53376/The-Dynamic-Two-Fluid-Model-OLGA-Theory-and
- Krasnopolsky & Lukyanov (2018), *Journal of Computational Physics* 355, 597–619: https://doi.org/10.1016/j.jcp.2017.11.032
- Public LedaFlow validation context: https://www.sintef.no/en/publications/publication/0198cc94bc36-07055a4c-7131-4f85-a956-faab6d1c2755/

Related: #2705. Keep that issue open until conservative terrain/slug source coupling and the complete acceptance matrix are implemented.

### DIFF
--- a/docs/process/TWOFLUIDPIPE_MODEL.md
+++ b/docs/process/TWOFLUIDPIPE_MODEL.md
@@ -386,6 +386,9 @@ There is no separate post-step projection toward the steady-state holdup correla
 projection changes phase inventory without a boundary flux or mass-transfer source and makes the
 error scale with pipe length. The earlier unreferenced 4 s relaxation time has therefore been
 removed; steady-state closures remain part of initialization and the local closure/source terms.
+Auxiliary terrain and slug trackers may maintain primitive diagnostics, but they do not rebuild the
+finite-volume phase masses. Conservative source or flux coupling for those trackers remains future
+model-development work.
 
 For a domain with no external mass source, validate the discrete balance
 
--- a/docs/process/TWOFLUIDPIPE_MODEL.md
+++ b/docs/process/TWOFLUIDPIPE_MODEL.md
@@ -381,6 +381,25 @@ root-mean-square pressure difference across all sections; for compact tests a li
 useful sanity check. For multiphase cases, also compare liquid, oil, and water holdup profiles so
 the dynamic calculation has settled hydraulically, not only acoustically.
 
+Transient holdup is reconstructed from the phase masses advanced by the finite-volume equations.
+There is no separate post-step projection toward the steady-state holdup correlation. Such a
+projection changes phase inventory without a boundary flux or mass-transfer source and makes the
+error scale with pipe length. The earlier unreferenced 4 s relaxation time has therefore been
+removed; steady-state closures remain part of initialization and the local closure/source terms.
+
+For a domain with no external mass source, validate the discrete balance
+
+$$
+M(t + \Delta t) - M(t) =
+\int_t^{t+\Delta t} \left(\dot m_{in} - \dot m_{out}\right)\,dt,
+\qquad
+M = \sum_i \left(m'_{g,i} + m'_{o,i} + m'_{w,i}\right)\Delta x_i .
+$$
+
+Use `getTotalMassInventory()` to read $M$ in kg directly from the conservative gas, oil, and water
+masses. A steady-state solution remains a useful long-time comparison, but agreement must result
+from the transient balances and closure forces rather than overwriting the conserved state.
+
 ### Boundary Conditions
 
 Use `setInletBoundaryCondition(...)` and `setOutletBoundaryCondition(...)` for explicit boundary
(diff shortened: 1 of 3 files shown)


EvenSol:
CI diagnosis for the five `TwoFluidPipeBoundaryConditionTest` failures:

The expected mass bound is physically valid and should not be relaxed. A 6 -> 12 kg/s inlet step over 1 ms permits O(0.01 kg) net boundary transfer, not the observed 1925.25 kg domain-volume-scaled inventory jump.

The remaining non-conservative path is after the finite-volume update: with slug tracking enabled, `runTransient` executes accumulation/slug tracking and then unconditionally calls `synchronizeConservativeStateWithPrimitiveState()`. That calls `TwoFluidSection.updateConservativeVariables()` for every cell and reconstructs `alpha*rho*A` mass from primitive holdup/density across the full domain, even when no external mass source or actual slug transfer occurred. The same state overwrite can explain why the pressure-step cases move away from their stationary targets.

Required correction: finite-volume conservative mass must remain authoritative after initialization. Primitive closure/slug/terrain synchronization may redistribute phase mass only with an explicitly conservative transfer; a no-op synchronization must not recreate section inventory. Re-run the focused five tests before considering any settling-time/tolerance adjustment. The branch has been restored to clean head `0fdd7f7`; no broad or unvalidated file rewrite is retained.

Issue #2698, opened 2026-07-30
Audit unnecessary high thermodynamic initialization levels


EvenSol:
## Motivation

NeqSim should use the lowest thermodynamic initialization level required by the quantities actually consumed. Higher levels add avoidable derivative work in flash and process-simulation hot paths.

## Rule

- `init(1)` is sufficient for fixed-T/P EOS state, cubic-root selection, Z/molar volume, fugacity coefficients, chemical potentials, phase-equilibrium Gibbs comparisons, TPD/stability work, and TP-flash iteration state.
- `init(2)` should be used only when caloric properties or first temperature derivatives are required.
- `init(3)` should be used only for explicitly required second composition derivatives/full derivative matrices.
- `initProperties()` is separate and should be used only when physical/transport properties are requested.

## Audit scope

Search production code for `init(2)` and `init(3)`, prioritizing iterative or frequently executed paths. Initial code-search candidates include process heat-exchanger classes, mixer implementations, transient/flow-node calculations, and flash utilities. Each candidate must be handled independently:

1. Trace every downstream property getter.
2. Verify the minimum required level from the relevant `Phase`, `PhaseEos`, component, or property implementation.
3. Add focused regression tests proving unchanged requested values and convergence.
4. Measure iterations, flash calls, allocations, or a stable benchmark metric where performance-sensitive.
5. Do not bulk-replace initialization calls or reduce a level without evidence.

This issue is an audit tracker, not authorization to change all call sites at once. Open focused PRs for validated improvements only.

EvenSol:
Mixer audit outcome (current `master` `6c9b6d8`): `Mixer.calcMixStreamEnthalpy()` only consumes `getEnthalpy()`, so `init(2)` is technically sufficient and `init(3)` composition derivatives are unused. I benchmarked an isolated 4-inlet mixer and the same mixer inside `ProcessSystem`, using 10-component gas and 20-component characterized-oil SRK cases. Each result used 300 warmups and 9 × 900 measured state-changing runs on OpenJDK 17 with fixed heap/G1.

Median baseline → `init(2)`:

- gas mixer: 4922.051 → 4906.954 µs/run (0.3%)
- oil mixer: 4539.232 → 4442.708 µs/run (2.1%)
- gas `ProcessSystem`: 5495.156 → 5570.463 µs/run (-1.4%)
- oil `ProcessSystem`: 5321.984 → 5290.433 µs/run (0.6%)

[...]

PR #2738, opened 2026-08-02
perf: avoid derivative init in heat exchanger entropy

EvenSol:
## Bottleneck

`HeatExchanger.getEntropyProduction(...)` reruns both inlet and outlet streams and then explicitly requests thermodynamic initialization level 3 on all four states. The method only consumes entropy, temperature, and duty. Entropy is a caloric property available at level 2; level 3 additionally computes composition-derivative matrices that this diagnostic never reads.

This path is reached directly and through both `ProcessSystem.getEntropyProduction(...)` and `ProcessModel.getEntropyProduction(...)`, including process safety/exergy reporting.

Refs #2698.

## Change

Use `init(2)` for the two inlet and two outlet states in the two-stream heat-exchanger entropy diagnostic.

The change is intentionally limited to `HeatExchanger`. Adjacent `Heater`, `Cooler`, `MultiStreamHeatExchanger`, and mass-balance diagnostics remain unchanged until each has independent profiling evidence.

## Reproducible benchmark

Java 17.0.19, G1, fixed 512 MiB heap. Each result used five fresh JVM forks. Within each fork:

- build independent baseline/candidate SRK exchanger cases;
- 50 alternating warmups;
- 9 alternating-order samples;
- 250 calls per sample;
- report the median per fork, then the median across forks;
- candidate is the exact four-call level-3 to level-2 replacement;
- entropy is checked after every scope.

13-component rich-gas case:

| Scope | Baseline median | Candidate median | Paired median gain |
|---|---:|---:|---:|
| direct `HeatExchanger` | 0.485181 ms | 0.463317 ms | 4.51% |
| `ProcessSystem` | 0.489873 ms | 0.454363 ms | 5.80% |
| `ProcessModel` | 0.488592 ms | 0.450288 ms | 7.79% |

The five-component direct case was noise-level (0.180813 vs 0.185726 ms; overlapping sample ranges), so no small-fluid speedup is claimed.

Maximum baseline/candidate entropy difference across all forks and scopes: `0.0 J/K`.

The benchmark used the public NeqSim 3.16.0 fat JAR for the runtime baseline and a subclass containing only this replacement. The current-master target method and file blob were verified unchanged before publication. Exact-head Maven validation is delegated to GitHub Actions because this workspace cannot resolve the Maven Enforcer plugin from Central.

## Regression and engineering validation

The regression uses a clone-stable initialization counter and an independent level-3 reference calculation. It verifies at the base case and a nearby +5 K hot-inlet point that:

- entropy matches the level-3 reference;
- all four states still receive level-2 caloric initialization;
- no level-3 derivative initialization is requested by this method;
- hot/cold energy balance closes;
- exchanger duty is preserved;
- both side mass flows are preserved.

No public API, EOS, mixing rule, flash algorithm, tolerance, convergence criterion, thread-safety contract, serialization field, or process result changes.

## Documentation impact

Documentation impact: none. This changes only an internal initialization depth for a diagnostic; its public API, result, units, assumptions, and recommended usage are unchanged. Existing thermodynamic-property documentation already identifies level 2 as the caloric-property level and level 3 as composition derivatives.

## Limitations

This optimization matters for richer multicomponent diagnostics and aggregated model reporting. It does not accelerate the exchanger solve itself, and no speedup is claimed for the five-component case.

### DIFF
--- a/src/main/java/neqsim/process/equipment/heatexchanger/HeatExchanger.java
+++ b/src/main/java/neqsim/process/equipment/heatexchanger/HeatExchanger.java
@@ -771,9 +771,9 @@ public class HeatExchanger extends Heater implements HeatExchangerInterface, Sta
     for (int i = 0; i < 2; i++) {
       UUID id = UUID.randomUUID();
       inStream[i].run(id);
-      inStream[i].getFluid().init(3);
+      inStream[i].getFluid().init(2);
       outStream[i].run(id);
-      outStream[i].getFluid().init(3);
+      outStream[i].getFluid().init(2);
       entrop += outStream[i].getThermoSystem().getEntropy(unit) - inStream[i].getThermoSystem().getEntropy(unit);
     }
 
--- a/src/test/java/neqsim/process/equipment/heatexchanger/HeatExchangerTest.java
+++ b/src/test/java/neqsim/process/equipment/heatexchanger/HeatExchangerTest.java
@@ -1,6 +1,9 @@
 package neqsim.process.equipment.heatexchanger;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import neqsim.process.equipment.separator.Separator;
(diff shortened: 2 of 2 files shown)


EvenSol:
Addressed Copilot’s clone-diagnostics finding at exact head `4287f29`: the initialization-tracking clone now throws an explicit `IllegalStateException` if `SystemSrkEos.clone()` returns null. The inline thread is resolved; exact-head quality gates are rerunning.

PR #2752, opened 2026-08-02
perf: avoid derivative init in cooler entropy

EvenSol:
## Bottleneck

`Cooler.getEntropyProduction(...)` reruns its inlet and outlet streams, then explicitly requests thermodynamic initialization level 3 on both states even though it only reads entropy. Entropy is a level-2 caloric property; level 3 additionally computes composition-derivative matrices that this diagnostic never consumes.

The method also calculates an unused heat-transfer value and writes it directly to `System.out` on every call. This adds allocation, formatting, synchronization, and environment-dependent I/O to direct diagnostics and to `ProcessSystem` / `ProcessModel` entropy aggregation.

This is independent of the merged two-stream exchanger optimization in #2738 and does not overlap the active ProcessModel/convergence or column-cache work.

Refs #2698.

## Change

- initialize the cooler inlet and outlet at level 2;
- remove the unused heat-transfer multiplication and unconditional console output;
- preserve the returned entropy calculation exactly.

No neighboring heater, exchanger, solve, or transport-property path is changed.

## Reproducible benchmark

OpenJDK 17.0.19, G1, fixed 512 MiB heap, public NeqSim 3.16.0 fat JAR. The current-master `Cooler.java` blob was inspected before the run. Five fresh JVM forks were used. Within each fork:

- build independent baseline and candidate SRK cooler cases;
- 50 alternating warmups;
- 9 alternating-order samples;
- 500 calls per sample;
- report the median in each fork and the paired median gain across forks;
- use a 13-component rich gas at 40 bara, cooled from 90 to 30 °C;
- route baseline stdout to `/dev/null`, making the logging result a conservative lower bound;
- compare entropy after every scope.

| Scope | Baseline median-of-fork-medians | Candidate median-of-fork-medians | Paired median gain |
|---|---:|---:|---:|
| direct `Cooler` | 0.305494 ms | 0.285994 ms | 5.80% |
| `ProcessSystem` | 0.307475 ms | 0.281892 ms | 8.32% |
| `ProcessModel` | 0.301595 ms | 0.279063 ms | 8.31% |

A nearby 95 °C inlet produced an 8.41% paired median direct gain. Maximum baseline/candidate entropy difference across all forks and scopes was `0.0 J/K`.

The five-component case was noisy (-16.8% to +15.1% by fork), so no small-fluid speedup is claimed. Real console-backed execution should benefit more than this deliberately stdout-discarded benchmark, but no unmeasured I/O gain is claimed.

## Regression and engineering validation

The added regression uses a clone-stable initialization counter and an independent level-3 reference calculation. At 90 °C and a nearby 95 °C inlet it verifies:

- entropy equals the level-3 reference;
- inlet and outlet still receive level-2 caloric initialization;
- no level-3 derivative initialization is requested by this diagnostic;
- duty remains the inlet/outlet enthalpy difference;
- mass flow and phase counts remain unchanged.

The five-fork benchmark also exercised direct, `ProcessSystem`, and `ProcessModel` repeated calls with exact entropy equality.

No public API, EOS, mixing rule, flash algorithm, tolerance, convergence criterion, serialized field, cloning contract, or thread-safety contract changes.

## Documentation impact

Documentation impact: none. This removes internal diagnostic overhead and an accidental console side effect; the public method, return value, units, assumptions, and recommended usage are unchanged. Existing thermodynamic-property documentation already identifies level 2 as caloric initialization and level 3 as composition derivatives.

## Limitations

This accelerates entropy/exergy diagnostics and aggregate reporting, not the cooler solve itself. No speedup is claimed for the noisy five-component benchmark. Repository Maven, Spotless, Java 8/21, Windows/Linux, Javadocs, and static-analysis gates are delegated to GitHub Actions because this workspace does not contain a Maven checkout.

### DIFF
--- a/src/main/java/neqsim/process/equipment/heatexchanger/Cooler.java
+++ b/src/main/java/neqsim/process/equipment/heatexchanger/Cooler.java
@@ -54,15 +54,11 @@ public class Cooler extends Heater {
   public double getEntropyProduction(String unit) {
     UUID id = UUID.randomUUID();
     inStream.run(id);
-    inStream.getFluid().init(3);
+    inStream.getFluid().init(2);
     getOutletStream().run(id);
-    getOutletStream().getFluid().init(3);
+    getOutletStream().getFluid().init(2);
 
-    double heatTransferEntropyProd = coolingMediumTemperature * getDuty();
-    System.out.println("heat entropy " + heatTransferEntropyProd);
-    double entrop = getOutletStream().getThermoSystem().getEntropy(unit) - inStream.getThermoSystem().getEntropy(unit);
-
-    return entrop;
+    return getOutletStream().getThermoSystem().getEntropy(unit) - inStream.getThermoSystem().getEntropy(unit);
   }
 
   /** {@inheritDoc} */
--- a/src/test/java/neqsim/process/equipment/heatexchanger/CoolerTest.java
+++ b/src/test/java/neqsim/process/equipment/heatexchanger/CoolerTest.java
@@ -201,23 +201,30 @@ class CoolerTest {
   }
 
   private static void assertMinimumEntropyInitialization(InitTrackingSystemSrkEos fluid, Cooler cooler) {
-    double expectedEntropy = referenceEntropyProduction(cooler, "J/K");
     double inletEnthalpy = cooler.getInletStream().getFluid().getEnthalpy();
     double outletEnthalpy = cooler.getOutletStream().getFluid().getEnthalpy();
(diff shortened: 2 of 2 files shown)


EvenSol:
Addressed Copilot's initialization-order finding in [`226421e`](https://github.com/equinor/neqsim/commit/226421e67817abb03d11bf33c013dc29da41bd42). The candidate call now runs immediately after resetting counters; its level-2/level-3 counts and phase state are captured before the independent level-3 reference is executed. This prevents pre-initialized level-3 state from masking a conditional derivative-init regression.

PR #2764, opened 2026-08-02
perf: avoid derivative init in multi-stream entropy

EvenSol:
## Purpose

`MultiStreamHeatExchanger.getEntropyProduction()` reruns every inlet and outlet and then requests level-3 thermodynamic initialization, although the diagnostic only consumes entropy (a caloric property). Level 3 also evaluates composition derivatives that are unused here.

This changes only those six per-call initializations (three inlet/outlet pairs in the representative case) from level 3 to level 2. It extends the initialization-level audit in #2698 without closing that tracker. `getMassBalance()` and all exchanger execution paths remain unchanged.

## Benchmark

Exact production-class overlays were compiled from `master` with Java 8 source/target compatibility. Five fresh JVM forks used OpenJDK 17, `-Xms512m -Xmx512m -XX:+UseG1GC`, two warmup blocks, and nine measured blocks of 300 state-changing calls per mode. The workload was a 13-component SRK rich gas with three streams (100 °C hot, 20 °C cold, 0 °C coldest), a 5 °C approach, and alternating hot-inlet perturbations of ±0.2 °C.

Paired median improvements across five forks:

| Path | Baseline median | Candidate median | Improvement |
|---|---:|---:|---:|
| Direct exchanger | 1.07458 ms/call | 1.02288 ms/call | 4.47% |
| `ProcessSystem` | 1.08352 ms/call | 1.03930 ms/call | 3.91% |
| `ProcessModel` | 1.05647 ms/call | 1.04166 ms/call | 2.72% |

Every baseline and candidate fork produced the identical entropy checksum `18245.547398123060`. One direct fork was noisy and regressed; all five `ProcessSystem` and all five `ProcessModel` pairs improved, so the claim is limited to paired medians for this workload.

## Accuracy and robustness

The regression test:

- compares the optimized result with an independent level-3 reference at 100 °C and a nearby 105 °C hot-inlet condition;
- verifies every inlet/outlet still receives caloric initialization and no level-3 initialization occurs in the optimized diagnostic;
- preserves inlet/outlet enthalpy, mass flow, phase count, per-stream mass balance, and exchanger heat balance;
- executes through a `ProcessSystem` twice to cover repeated-run behavior.

There are no mutable caches, shared static state, API changes, or tolerance changes. Entropy values, duty, units, phase behavior, and process convergence are unchanged.

## Validation

- Java 8 source/target compilation of the modified production dependency closure: passed
- Java 8 source/target compilation of the focused test: passed
- Standalone execution of the regression invariants at both operating points: passed
- `git diff --check`: passed
- Full Maven/JUnit, Spotless, Javadocs, CodeQL, pre-commit, and PaperLab checks: delegated to CI

## Limitations

The measured workload is a 13-component SRK rich gas repeatedly invoking an entropy diagnostic. No performance claim is made for every EOS, component count, or ordinary exchanger execution that does not request entropy. The adjacent mass-balance diagnostic remains intentionally untouched.

Documentation impact: none. Public API, returned values, units, defaults, and recommended usage are unchanged; only an internal initialization depth is reduced.

### DIFF
--- a/src/main/java/neqsim/process/equipment/heatexchanger/MultiStreamHeatExchanger.java
+++ b/src/main/java/neqsim/process/equipment/heatexchanger/MultiStreamHeatExchanger.java
@@ -285,9 +285,9 @@ public class MultiStreamHeatExchanger extends Heater implements MultiStreamHeatE
     for (int i = 0; i < inStreams.size(); i++) {
       UUID id = UUID.randomUUID();
       inStreams.get(i).run(id);
-      inStreams.get(i).getFluid().init(3);
+      inStreams.get(i).getFluid().init(2);
       outStreams.get(i).run(id);
-      outStreams.get(i).getFluid().init(3);
+      outStreams.get(i).getFluid().init(2);
       entropyProduction += outStreams.get(i).getThermoSystem().getEntropy(unit)
           - inStreams.get(i).getThermoSystem().getEntropy(unit);
     }
--- a/src/test/java/neqsim/process/equipment/heatexchanger/MultiStreamHeatExchangerTest.java
+++ b/src/test/java/neqsim/process/equipment/heatexchanger/MultiStreamHeatExchangerTest.java
@@ -297,8 +297,7 @@ public class MultiStreamHeatExchangerTest {
     int actualLevelThreeCalls = fluid.getLevelThreeCalls();
 
     assertEquals(expectedEntropy, actualEntropy, Math.max(1.0e-10, Math.abs(expectedEntropy) * 1.0e-12));
-    assertTrue(actualLevelTwoCalls >= 2 * streamCount,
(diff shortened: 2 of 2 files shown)


EvenSol:
CI status on exact head `71cda400`:

- Spotless format patch: passed after applying the formatter's two wrapping changes to the new test.
- PaperLab replication gate: passed.
- Repository-wide pre-commit: still fails only because current `master` contains pre-existing Spotless violations in `TPmultiflash.java` and `TPmultiflashPhaseDisappearanceTest.java`. Neither file is touched by this PR. [Run 30763991390](https://github.com/equinor/neqsim/actions/runs/30763991390)
- Java build/tests/Javadocs and CodeQL are still running.

No unrelated formatting files were added to this focused performance PR.

PR #2793, opened 2026-08-04
fix(tpflash): clear stale beta-solver stall state

EvenSol:
## Problem

Merged #2767 stores `betaSolveStalled` on the `TPmultiflash` operation, but only assigns it inside the `multiPhaseTest` beta-solver block. Re-entering or reusing the same operation for a pass where that block does not execute can therefore retain a previous `true` value. The later three-phase rescue reads the field unconditionally.

## Change

- reset `betaSolveStalled` at the start of every `run()`;
- add a focused reuse regression that plants a prior stall state, executes a single-phase pass without a beta solve, and requires the state to be cleared.

No flash equation, phase-selection rule, tolerance, public API, or successful beta-solver result changes.

## Validation

Hosted CI will run the focused and complete repository gates. The regression is intentionally state-lifecycle focused and fails on current master because the planted value remains true.

## Copilot disposition

This directly addresses the unresolved stale-`betaSolveStalled` thread on merged #2767. All new Copilot feedback on this PR must be assessed and validated before it is marked ready.

Documentation impact: none — this restores private per-run solver-state ownership and does not change documented thermodynamic behavior or API usage.

Follow-up to #2767.

### DIFF
--- a/src/main/java/neqsim/thermodynamicoperations/flashops/TPmultiflash.java
+++ b/src/main/java/neqsim/thermodynamicoperations/flashops/TPmultiflash.java
@@ -2125,6 +2125,7 @@ public class TPmultiflash extends TPflash {
   public void run() {
     int aqueousPhaseNumber = 0;
     enhancedStabilityChecked = false;
+    betaSolveStalled = false;
     // logger.info("Starting multiphase-flash....");
 
     // For systems with ions, temporarily remove ions before stability analysis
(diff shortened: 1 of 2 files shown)


EvenSol (commented):
Maintainer review: **ready to merge from a code-review perspective**.

The reset belongs at the start of `TPmultiflash.run()`: it preserves successful beta-solver behavior while preventing a stale `betaSolveStalled=true` value from influencing a later reused-operation pass where the beta solver does not execute. The focused regression directly covers that reuse path. No flash equations, convergence criteria, phase-selection logic, or public API are changed.

All reported workflows on head `03e164b4a0f9ea8397dfdfc3f7aa0f65a1f18ba8` pass, including build/test/Javadoc, pre-commit, Spotless, CodeQL, and PaperLab.

I found no blocking issue. GitHub does not allow this account to formally approve its own PR, so this is recorded as a non-blocking review comment instead.

PR #2807, opened 2026-08-05
fix(column): honor active gates in exact warm-state reuse

EvenSol:
## Problem

An accepted `NAPHTALI_SANDHOLM` result could receive an exact zero-iteration cache hit after a caller changed an active convergence tolerance. The physical input fingerprint remained identical and the prior solve status was still accepted, so `canReuseNaphtaliSandholmWarmState(...)` returned true even though the current invocation had a different convergence contract.

That could return a state without giving the solver an opportunity to meet the new contract or report fresh non-convergence.

## Reproduction and root cause

The deterministic six-stage SRK propane/n-butane/n-pentane stripper already used by the warm-state regression suite is solved with `NAPHTALI_SANDHOLM`, rerun unchanged to establish legitimate exact reuse, and then given a MESH tolerance immediately below its accepted finite residual.

On unmodified `master` `1458ae72550fe3928954c4f2080d0b777f573ec3`:

- `solved()` immediately reflects that the stored state is outside the tightened gate;
- the process-input fingerprint still matches;
- `willReuseNaphtaliSandholmWarmState()` nevertheless predicts a cache hit;
- the next invocation can return that state without executing the solver path.

A standalone packaged-runtime probe showed the same class of defect when the mass-balance tolerance was tightened below the stored residual: the invocation reported zero iterations while the stored state no longer met the newly requested gate.

The root cause is that the exact cache recorded the process-input fingerprint but not the convergence-gate configuration under which the complete public solve was accepted.

## Change

Record a deterministic convergence-contract fingerprint after the complete public column solve and require it to match before Naphtali-Sandholm exact reuse.

The fingerprint covers:

- effective temperature, mass, and enthalpy tolerances;
- energy-gate enforcement;
- effective MESH-gate enforcement and tolerance;
- tray-material and terminal product-draw tolerances;
- column-tear and pumparound tolerances.

The snapshot is deliberately committed after coordinated solver/fallback and product-reconciliation handling. It therefore does not depend on mutable intermediate residual diagnostics.

No tray equations, thermodynamic flashes, damping, default tolerances, iteration limits, product reconciliation, or fallback algorithms change.

## Validation

The new regression checks:

- a finite positive MESH residual accepted by the original tolerance;
- unchanged-input zero-iteration reuse remains available;
- tightening the MESH gate makes the existing state non-converged;
- exact-reuse prediction is rejected under the changed contract;
- the next run is not labelled as a reused warm state;
- total/component balance and physical product bounds remain valid.

The first full CI round also exposed an important boundary in an existing component-identity warm-state regression: gating directly on mutable residual diagnostics blocked legitimate reuse after wrapper reconciliation. That design was removed. The current implementation uses only the post-public-solve contract snapshot and is being revalidated on exact head.

## Compatibility and risk

The observable change is limited to callers that change an active convergence-gate configuration after an accepted Naphtali-Sandholm solve. Such calls execute the solver path instead of returning a cache hit under a different contract. Unchanged configurations keep the existing zero-iteration performance. All API signatures and default tolerances remain unchanged.

Remaining limitation: changing a gate does not make a tighter target feasible; it ensures the result is freshly attempted and its convergence status is honest.

## Documentation impact

The distillation equipment guide and agent changelog document that exact reuse requires both an identical process fingerprint and the same post-solve convergence-gate configuration. No notebook changed.

### DIFF
--- a/CHANGELOG_AGENT_NOTES.md
+++ b/CHANGELOG_AGENT_NOTES.md
@@ -13,10 +13,10 @@
 
 ### Corrected
 
-Naphtali-Sandholm exact unchanged-input reuse now requires the stored result to satisfy the
-convergence gates that are active for the current invocation. Tightening an enforced tolerance no
-longer returns the previously accepted state with zero iterations when that state is outside the
-new contract. Unchanged or loosened gates retain the existing reuse performance.
+Naphtali-Sandholm exact unchanged-input reuse now requires the active convergence-gate
+configuration to match the snapshot recorded after the accepted public solve. Changing an enforced
+tolerance no longer returns the previous state with zero iterations under a different convergence
+contract. Unchanged gates retain the existing reuse performance.
 
 ## 2026-08-04 — TwoFluidPipe closed thermal boundary consistency
 
--- a/CHANGELOG_AGENT_NOTES.md
+++ b/CHANGELOG_AGENT_NOTES.md
@@ -18,6 +18,21 @@ configuration to match the snapshot recorded after the accepted public solve. Ch
 tolerance no longer returns the previous state with zero iterations under a different convergence
 contract. Unchanged gates retain the existing reuse performance.
 
+## 2026-08-04 — Fixed-reflux fallback product inventory
+
+### Corrected
+
+When a fixed-liquid-reflux column rejects its tray state and installs guarded full-feed fallback
+products, the condenser's separate liquid product is now cleared. The fallback gas and bottom
+streams already contain the complete feed inventory; retaining the rejected liquid product beside
+them previously exposed more material than entered the column. Fixed-reflux availability,
+delivery, and residual diagnostics are invalidated because they no longer describe the exposed
+fallback products.
+
+The solve status remains `FALLBACK_PRODUCTS`, `solved()` remains false, and no tray equation,
+tolerance, flash, or iteration rule changed. Callers must continue to inspect the solve status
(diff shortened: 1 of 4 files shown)


EvenSol:
Copilot disposition on latest pushed head `8ee4f8f410aea8d765cc8942a34e30ccc8f873c9`:

- Accepted the suggestion to avoid coupling the regression to human-readable status text; the `getLastSolveStatusReason().contains("Reused")` assertion was removed. The test now uses the stable programmatic reuse flags and iteration count.
- Rejected the suggested direct `residualConvergenceSatisfied()` reuse gate for this bounded fix. That was the first implementation and the exact-head full CI round demonstrated a regression in the existing `componentIdentityChangeForcesColdInitialization` case: public wrapper/product reconciliation can legitimately mutate the residual telemetry after the simultaneous solver returns, so those mutable diagnostics are not a sound cache-validity snapshot.
- Retained exact convergence-configuration equality intentionally. It is conservative: any changed gate executes the solver path, while unchanged gates retain zero-iteration reuse. Allowing loosened or inactive-gate changes safely would require immutable post-public-solve residual snapshots and a compatibility relation; that is a separate optimization, not a correctness blocker for this PR.
- Documentation and changelog now state that exact-configuration contract explicitly, so they match the implemented behavior.

There are no inline review threads. The refined implementation still requires a fresh exact-head CI run after the branch is brought up to date with current `master`; the PR is not being represented as ready.

PR #2863, opened 2026-08-06
Fix separator liquid outlet phase misclassification after restore-like runs

Sviatose:
## Summary
This PR fixes a separator phase-consistency bug where a liquid outlet stream could be unnecessarily TP-reflashed and, in edge cases, become misclassified as gas-like even for heavy reflux compositions.

## What happened
In the separator `run(UUID id)` path, the liquid outlet stream was always re-run when an oil/aqueous phase existed (and component count was > 1), regardless of whether gas-in-liquid entrainment correction was requested.

That extra outlet `run()` call is not required in the normal case (`gasInLiquid == 0.0`) because the outlet fluid has already been taken from the flashed separator phase (`setThermoSystemFromPhase(..., "liquid")`). Reflashing that isolated outlet can alter phase classification in restore-like edge states.

## Root cause
The gating logic for liquid-outlet reflash was too broad:
- It triggered on phase presence + component count
- It did **not** require an actual gas-in-liquid entrainment correction

## Fix
In `Separator.run(UUID id)`:
- Keep gas outlet behavior unchanged
- Restrict `liquidOutStream.run(id)` to only cases where it is actually needed:
  - oil/aqueous phase exists
  - component count > 1
  - `gasInLiquid != 0.0`
- Otherwise finalize the already phase-selected outlet using `finalizePhaseOutlet(...)`

This preserves the flashed separator phase identity for normal operation and still performs reflash when entrainment correction demands it.

## Regression test
Added `SeparatorPhaseConsistencyTest`:
- Programmatically builds a heavy reflux-like feed (no XML fixture dependency)
- Reproduces the risky path with an existing liquid outlet stream object
- Verifies no unnecessary liquid outlet `run()` occurs when `gasInLiquid == 0.0`
- Verifies outlet remains liquid-like (not gas-dominant) and composition remains heavy

## Why we need this
Without this guard, valid heavy liquid outlets can be destabilized by an avoidable extra flash step, potentially triggering downstream guard trips and incorrect behavior in reflux/pump handling.

This change narrows behavior to the physically intended cases, improves restore robustness, and adds focused regression coverage to prevent recurrence.

## Scope and risk
- Small, targeted logic change in one conditional
- No API changes
- Existing behavior for entrainment-correction scenarios preserved
- Verified with targeted test: `SeparatorPhaseConsistencyTest`


EvenSol (approved):
good fix

PR #2864, opened 2026-08-06
Fix compressor PS/PH NaN robustness and sanitize deserialized efficiencies

Sviatose:
## Summary
This PR fixes a compressor failure mode where pathological/restored states trigger NaN in the flash path (notably in `PSflash` via EOS root selection), and hardens efficiency handling for deserialized artifacts.

## Problem
A replay of the failing compressor artifact showed:
- Restored efficiencies above physical bounds (`eta_isen > 1`, `eta_poly > 1`)
- `PSflash` failure with NaN (`PhaseSrkEos:molarVolumeAnalytical - compressibility factor is NaN`)

Even with setter-level clamping, deserialization can populate fields directly and bypass setters.

## Root Cause
1. **Deserialized state bypassed efficiency setters**
   - Invalid efficiency values survived into runtime and destabilized downstream calculations.
2. **`PSflash` path lacked deterministic NaN recovery**
   - When `PSflash` entered a numerically fragile region, compressor run aborted instead of recovering.
3. **Efficiency back-calculation secant loop became singular under bounded updates**
   - After introducing strict bounds, the secant slope could collapse and produce NaN during `solveEfficiency` unless stabilized.

### 1) Runtime sanitization of restored efficiencies
- Added run-start sanitization to enforce `0 < eta <= 1` even for deserialized objects:
  - `sanitizeEfficiencyStateForRun()` called at start of `run(UUID)`.

### 2) Robust PS flash recovery in compressor path
- Hardened `runRegularPsFlash(...)`:
  - Catches NaN-related runtime failures in `PSflash`.
  - Falls back deterministically to TP-based entropy solve (`S(T,P)=S_target`) using existing robust helper.
  - Also falls back when post-PS entropy is non-finite or residual is outside tolerance.

### 3) Stability fix for bounded efficiency solver loop
- Updated `solveEfficiency(...)` secant loop to handle near-zero/non-finite slope safely:
  - Directional fallback step if secant derivative is singular.
  - Non-finite guard on candidate efficiency.
  - Denominator protection for next iteration.
  - Returns bounded current efficiency.

### 4) Regression tests
- Added focused test class:
  - `CompressorEfficiencyBoundsTest`
  - Verifies setter clamping and deserialize-like direct field assignment sanitized at run start.

## Validation
Executed successfully:
- `./mvnw -Dtest=neqsim.process.equipment.compressor.CompressorEfficiencyBoundsTest test -DskipITs`
- `./mvnw -Dtest=neqsim.process.equipment.compressor.ASMEPTC10ValidationTest,neqsim.process.equipment.compressor.CompressorEfficiencyBoundsTest test -DskipITs`

Also replayed the failing compressor artifact after fix:
- Before run: `eta_isen=2.446...`, `eta_poly=2.219...`
- After run: bounded and converged (`eta_isen=1.0`, finite `eta_poly`), no NaN crash, finite outlet T/P.

## Why this is safe
- Keeps physical constraints explicit (`0 < eta <= 1`).
- Uses thermodynamically consistent fallback at fixed pressure by solving `S(T,P)=S_target` via TP flashes.
- Preserves normal behavior (PTC10 validation remains green) while fixing pathological-state robustness.

### DIFF
--- a/src/main/java/neqsim/process/equipment/compressor/Compressor.java
+++ b/src/main/java/neqsim/process/equipment/compressor/Compressor.java
@@ -4625,7 +4625,8 @@ public class Compressor extends TwoPortEquipment
       }
       cleanEfficiency = designPolytropicEfficiency;
     }
-    setPolytropicEfficiency(cleanEfficiency * effMultiplier);
+    // Preserve the clean baseline so repeated run() calls do not compound the degradation.
+    polytropicEfficiency = normalizeEfficiency(cleanEfficiency * effMultiplier, "polytropicEfficiency");
   }
 
   /**
--- a/src/main/java/neqsim/process/equipment/compressor/Compressor.java
+++ b/src/main/java/neqsim/process/equipment/compressor/Compressor.java
@@ -65,6 +65,12 @@ public class Compressor extends TwoPortEquipment
   /** Maximum number of TP-flash Newton iterations in the compressor entropy fallback. */
   private static final int MAX_ENTROPY_FLASH_ITERATIONS = 30;
 
+  /** Maximum physically meaningful compressor efficiency. */
+  private static final double MAX_EFFICIENCY = 1.0;
+
+  /** Small positive floor used when efficiency is configured as zero/negative/non-finite. */
+  private static final double MIN_EFFICIENCY = 1.0e-6;
+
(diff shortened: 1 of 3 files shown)


EvenSol (approved):
good update

PR #2881, opened 2026-08-07
Validate severe-slugging diagnostics against public experiments

### LINKED ISSUE #2741: TwoFluidPipe severe-slugging diagnostic omits flowline-riser system physics
## Problem

On current `master` ([`00ff56bc0f0e97b79bb510d06f2fdbe7b003309c`](https://github.com/equinor/neqsim/commit/00ff56bc0f0e97b79bb510d06f2fdbe7b003309c)), the public severe-slugging diagnostic is labeled as a Bøe/Pots-style stability number but is only a local inclined-cell Froude/holdup screen.

[`TwoFluidConservationEquations.calcSevereSluggingNumber(...)`](https://github.com/equinor/neqsim/blob/00ff56bc0f0e97b79bb510d06f2fdbe7b003309c/src/main/java/neqsim/process/equipment/pipeline/twophasepipe/TwoFluidConservationEquations.java#L433-L448) calculates

```
N_local = |v_sG| / (alpha_L * sqrt(g D (rho_L - rho_G) / rho_L))
```

for every section inclined more than 5 degrees and flags `N_local < 1` as severe-slug potential. The public [`getSevereSluggingNumberProfile()`](https://github.com/equinor/neqsim/blob/00ff56bc0f0e97b79bb510d06f2fdbe7b003309c/src/main/java/neqsim/process/equipment/pipeline/TwoFluidPipe.java#L4850-L4868) describes this as a severe-slugging stability number.

A second terrain path says it uses the “Pots severe slugging criterion,” but actually applies another local critical gas-velocity expression and an empirical holdup multiplier ([`TwoFluidPipe.java` lines 2842–2858](https://github.com/equinor/neqsim/blob/00ff56bc0f0e97b79bb510d06f2fdbe7b003309c/src/main/java/neqsim/process/equipment/pipeline/TwoFluidPipe.java#L2842-L2858)). Neither path evaluates a flowline-riser system stability criterion.

This is not covered by an open issue or PR. It is distinct from the conservative handoff, transient balance, and phase-disappearance fixes in #2724, #2729, and #2735.

## Reproducible analytical baseline

Use the current local equation with SI inputs:

- gas density: `rho_G = 30 kg/m3`
- liquid density: `rho_L = 800 kg/m3`
- diameter: `D = 0.1016 m`
- liquid holdup: `alpha_L = 0.5`
- superficial gas velocity: `v_sG = 1.0 m/s`
- gravity: `g = 9.81 m/s2`

The current result is

```
N_local = 2.041964882717
```

and is identical for both of these physically different systems when the local section state is held fixed:

| Case | Upstream flowline | Riser height | Separator pressure | Current result |
|---|---:|---:|---:|---:|
| compact/low backpressure | 100 m | 10 m | 1 bara | 2.041964882717 |
| long/deep/high backpressure | 10 km | 300 m | 20 bara | 2.041964882717 |

The implementation cannot respond directly to upstream gas volume/compressibility, total riser height, separator pressure, or choke characteristic because none is an argument. Yet these are defining variables in published pipeline-riser stability analyses.

A regression should first encode this invariance and fail the expectation that the public *system stability* diagnostic distinguishes the two systems. Do not tune the local threshold to make one case pass.

## Existing validation does not validate severe slugging

[`TwoFluidPipeValidationTest.testSevereSlugConditions()`](https://github.com/equinor/neqsim/blob/00ff56bc0f0e97b79bb510d06f2fdbe7b003309c/src/test/java/neqsim/process/equipment/pipeline/twophasepipe/TwoFluidPipeValidationTest.java#L490-L556):

- runs only `run()` (steady state);
- does not call `runTransient(...)`;
- does not read the stability-number or risk profiles;
- does not calculate a Bøe, Pots, or Taitel criterion;
- does not compare with published measurements;
- asserts only that riser-base holdup is positive.

The test class nevertheless says it validates “published OLGA validation cases” and “terrain-induced slugging field data patterns,” and the comparison guide reports Lagrangian slug tracking as validated. Those claims exceed the executable evidence.

## Physical basis

Severe slugging is a global instability of a downward/horizontal flowline feeding a rising section. The relevant competition is between compressed-gas pressure buildup upstream of the liquid blockage and the hydrostatic/backpressure response of the riser/top boundary.

- Taitel (1986) derives a pipeline-riser stability condition from the force derivative as gas penetrates the riser and shows explicit separator-pressure dependence: https://doi.org/10.1016/0301-9322(86)90026-1
- Jansen, Shoham & Taitel (1996) extend the stability analysis to choking and gas lift and explicitly state that pipeline length represents pipeline gas volume: https://doi.org/10.1016/0301-9322(96)00027-4
- Pots, Bromilow & Konijn (1987) report small-scale flowline/riser experiments and define severe slugs relative to riser length: https://doi.org/10.2118/13723-PA
- Chang et al. (2024) provide a recent long-distance pipeline-riser experimental/prediction study suitable for a modern validation target: https://doi.org/10.1016/j.oceaneng.2024.117636

These references guide a clean-room NeqSim implementation only. They do not justify copying proprietary OLGA/LedaFlow algorithms or claiming simulator equivalence. Public LedaFlow material identifies slug management and startup/shutdown/restart as transient use cases, but does not expose proprietary closures: https://falkor.com/products/ledaflow

## Engineering impact

The current API can:

- label an ordinary isolated uphill section as having system-level severe-slug risk;
- miss the effect of upstream gas volume, riser height, separator pressure, or choking;
- produce the same reported “stability number” for systems whose published stability models differ;
- overwrite/compete with the separate terrain-accumulation risk flag;
- give users false confidence because a steady positive-holdup test is presented as severe-slug validation.

This affects riser design, turndown/minimum stable flow, late-life operation, startup, shutdown, and slug-control studies.

## Proposed bounded implementation sequence

Do **not** replace the current formula with another local constant or silently change the meaning of the public getter.

1. Retain the present local quantity, if useful, under an honest name such as `inclinedSectionGasCarryoverNumber`; preserve the old API through a documented deprecated alias if needed.
2. Add a pipeline-riser system descriptor that identifies:
   - flowline gas volume and inclination;
   - riser-base location and total vertical rise;
   - riser holdup/void-fraction model;
   - top boundary/separator pressure;
   - optional choke pressure-flow characteristic;
   - gas compressibility/model assumptions.
3. Implement one published, dimensionally explicit screening criterion (Bøe/Taitel first) as a diagnostic, with assumptions and validity range. Keep it separate from the fully dynamic solution.
4. Validate the dynamic model independently; do not force transient results to match the diagnostic.

## Acceptance criteria

- [ ] A test proves the current analytical baseline and demonstrates that a system-level diagnostic changes when flowline gas volume, riser height, separator pressure, or choke response changes.
- [ ] The selected Bøe/Taitel/Pots equations, symbols, SI units, assumptions, and parameter provenance are documented; no unsupported formula is attributed to a source.
- [ ] A local carryover/Froude screen is not reported as a pipeline-riser severe-slug stability criterion.
- [ ] `getSevereSluggingNumberProfile()` has a backward-compatible migration path and unambiguous semantics.
- [ ] Detection is restricted to a valid flowline-low-point-riser topology and required flow regime, rather than every section above 5 degrees.
- [ ] A public experimental benchmark is encoded with source, geometry, fluids, pressure, temperature, boundary conditions, and digitization uncertainty. Start with Schmidt/Pots/Taitel or the public Park S-riser cases; add Chang et al. if the full dataset can be reproduced legally.
- [ ] The benchmark checks stable/unstable classification and, for dynamic cases, riser-base pressure period/amplitude, liquid-production cycling, maximum slug length relative to riser height, and phase inventories.
- [ ] Sweep nearby gas/liquid superficial velocities across the published transition boundary; report confusion matrix or boundary error rather than one tuned point.
- [ ] Demonstrate timestep and mesh convergence and phase-resolved/total mass closure using the diagnostics added by #2729.
- [ ] Cover repeated runs, copy/serialization, deterministic results, and unchanged behavior when slug diagnostics are disabled.
- [ ] Cover one-phase rejection, gas-liquid operation, and document three-phase limitations rather than implying validated oil-water-gas severe-slug prediction.
- [ ] Replace unsupported “OLGA-style,” “OLGA validation,” “OLGA-equivalent,” and LedaFlow-equivalence wording in affected Javadocs, tests, and guides with evidence-specific language.
- [ ] Preserve public API compatibility unless a migration is explicitly justified and documented.

## Updated prioritized two-fluid gap map

1. **P1 — severe-slug diagnostic correctness and topology:** this issue.
2. **P1 — steady/dynamic phase-transition consistency:** dew-point and inlet-ramp cases after #2724/#2729/#2735.
3. **P2 — public transient benchmark ladder:** pressure/holdup steps and stable/unstable flowline-riser boundaries.
4. **P2 — hydrodynamic and severe slug dynamics:** growth, period, amplitude, slug frequency/length, and riser cycles after the diagnostic is corrected.
5. **P2 — three-phase coupling:** oil-water slip, inversion/dispersion, wetting, gas entrainment, and independent phase transitions.
6. **P2 — startup/shutdown and minimum stable flow:** after topology, phase degeneracy, and transient benchmarks are reliable.
7. **P3 — heat/mass-transfer benchmarks and performance:** only after conservative hydrodynamics and stability validation.

## Documentation impact

No code changed in this issue-only step. The eventual fix is user-visible and must update `TwoFluidPipe` and `TwoFluidSection` Javadocs, `docs/process/TWOFLUIDPIPE_MODEL.md`, `docs/wiki/two_fluid_model_olga_comparison.md`, the reporting/validation guide, validation-test names/comments, and any examples that present the local number as a severe-slug stability result.

EvenSol:
## Summary

- add a reusable severe-slugging classifier benchmark harness and all 55 operating points from Tengesdal's public 2002 -3° air/mineral-oil flow map, including tabulated velocity uncertainty and explicit figure-digitization provenance
- add a slow RK4 reproduction of large-facility Test 3 as a **four-member ensemble**, covering mass closure, time-averaged riser-base pressure, the blowout/fallback regime signature, pressure amplitude, cycle period, tracked slug length, mesh/outer-step/perturbation sensitivity, and repeated-run determinism
- expose an explicit random seed for stochastic Lagrangian slug initiation so validation runs can be reproducible without changing the stochastic default
- expose `TwoFluidPipe.isSteadyStateWallClockLimited()` and `getSteadyStateIterationsUsed()` so a wall-clock-truncated, machine-speed-dependent steady-state initialization can no longer pass unnoticed
- replace unsupported commercial-simulator parity wording and disable two synthetic tests that could be mistaken for experimental severe-slugging evidence
- document evidence levels, assumptions, metrics that agree, and known residuals

## Why the benchmark had to be restructured

The first revision asserted that the 12- and 16-section meshes agree on peak-to-peak riser-base pressure within 20%. That passed locally and failed on CI at 107.0 kPa vs 43.5 kPa.

The cause is not a bug and not machine speed: **severe slugging in this configuration is a deterministically chaotic limit cycle.** Measured on 10 realizations, a relative inlet-pressure perturbation of 1e-12 — twelve orders of magnitude below the digitization uncertainty of the source figure — moves the instantaneous metrics by:

| Metric | Range over 10 realizations | Spread |
|---|---|---|
| Peak-to-peak riser-base pressure | 41.8 – 299.6 kPa | factor 7.2 |
| Apparent cycle period | 14.5 – 35.3 s | factor 2.4 |
| Maximum tracked outlet slug | 1.52 – 4.90 m | factor 3.2 |
| **Time-averaged riser-base pressure** | **170.7 – 176.5 kPa** | **±1.7%** |
| **Phase-resolved mass closure** | **7e-16 – 1.1e-15** | — |

A single trajectory is therefore not reproducible across platforms, compilers or JIT states. Asserting numerical agreement on instantaneous extremes produces a test that passes or fails by luck, and the previously reported "107 kPa / 21.8 s" were one lucky draw rather than a model property.

## Restructured dynamic benchmark

The benchmark now evaluates four realizations — 12 sections at 0.1 s, the same case with a 1e-12 inlet perturbation, 16 sections at 0.1 s, and 12 sections at 0.2 s — and separates trajectory-robust from trajectory-sensitive quantities.

**Asserted directly (trajectory-robust):**

- phase-resolved and total mass closure below 1e-10 (observed ~1e-15)
- the time-averaged riser-base pressure agrees within 8% across mesh, outer step and perturbation (observed spread below 4%)
- every realization shows the severe-slugging signature: outlet liquid rate blows out above 1.25× and falls back below 0.75× the liquid feed rate
- the steady-state initialization did not hit the wall-clock guard

**Bounded by physics only (trajectory-sensitive), reported as an ensemble range:**

- peak-to-peak pressure inside 0.2–4.0 riser hydrostatic heads (ρ_L g h = 125.4 kPa); observed 0.33–2.39
- the ensemble range must bracket the digitized 98 ± 5 kPa; observed 42–300 kPa
- each realization's period above the riser filling time, and the **ensemble-mean** period asserted to stay below the measured 38 ± 2 s (observed mean ≈ 23.6 s)
- maximum tracked slug positive and below one riser height; observed 0.10–0.33 riser heights

## Machine-speed safety

Independently of the chaos issue, `runSteadyState()` had a genuine machine-speed dependency: the wall-clock guard truncates the refinement loop, so a slow or loaded runner silently hands the transient a different initial condition. The benchmark now disables the guard (`setSteadyStateMaxWallClockTime(Double.POSITIVE_INFINITY)`; the loop is still bounded by its own 100-iteration cap) and asserts `isSteadyStateWallClockLimited()` is false for every realization. The two new accessors let any user detect the same silent truncation.

## Conclusions that stand

The model reproduces the severe-slugging *regime* — cyclic blowout and fallback with a riser-base pressure swing of the same order as the riser hydrostatic head, bracketing the digitized experimental amplitude — with exact mass conservation and a mesh/step-converged mean pressure. The cycle period is systematically short and the outlet tracker underpredicts the experimental severe-slug length. This PR therefore still does **not** claim quantitative severe-slugging validation. The missing explicit upstream tank/plenum, nitrogen/Crystex surrogate choices, 25 °C assumption, coarse-grid cell-placement sensitivity, and RK4-only coverage remain documented.

### 55-point velocity map (unchanged)

Transition observations are reported separately rather than forced into a binary class.

| Experimental class | Predicted severe | Predicted stable |
|---|---:|---:|
| Severe slug | 22 | 4 |
| Stable | 8 | 7 |
| Transition (not scored) | 6 | 8 |

Binary accuracy is 70.7%; severe-slug recall is 84.6%; stable recall is 46.7%. This supports use as a conservative screen, not as a high-specificity dynamic classifier.

Source: [Tengesdal (2002), BSEE Technical Assessment Program](https://www.bsee.gov/sites/bsee.gov/files/tap-technical-assessment-program/397aa.pdf)

## Validation

- `SevereSlugging*Test` + `LagrangianSlugTrackerTest` with slow tests enabled: 33 tests passed
- plus `TwoFluidPipeBoundaryConditionTest`, `TwoFluidPipeMassBalanceTest`, `TwoFluidPipePhaseDegeneracyTest`, `TwoFluidVsBeggsBrillComparisonTest`: 106 tests passed, 3 skipped
- `mvnw spotless:apply` and `spotless:check`: passed
- `mvnw javadoc:javadoc`: passed, no errors or warnings
- pre-commit and pre-push Spotless and documentation-search hooks: passed

Closes #2741


Sviatose:
Can we use it to test https://github.com/equinor/neqsim/blob/agent/severe-slugging-experimental-benchmark/src/main/java/neqsim/process/measurementdevice/simpleflowregime/SevereSlugAnalyser.java ?

EvenSol:
> Can we use it to test https://github.com/equinor/neqsim/blob/agent/severe-slugging-experimental-benchmark/src/main/java/neqsim/process/measurementdevice/simpleflowregime/SevereSlugAnalyser.java ?

Yes - we should do.
Even

EvenSol:
The CI failure (`mesh pressure amplitudes=107042.1 and 43524.6`) was not a bug and not a slow-machine timeout — it was the benchmark asserting a **non-reproducible quantity**.

I measured 10 realizations of the Test 3 case on this branch. Perturbing the inlet pressure by a relative `1e-12` (twelve orders of magnitude below the figure-digitization uncertainty) gives:

```
n=12 pert=0e+00 dt=0.1 | period=21.80 slug=1.923 meanP=172168 p2p=107042
n=12 pert=1e-12 dt=0.1 | period=35.30 slug=4.903 meanP=173596 p2p= 62979
n=12 pert=1e-09 dt=0.1 | period=25.80 slug=1.595 meanP=175058 p2p=299567
n=12 pert=1e-06 dt=0.1 | period=25.50 slug=1.521 meanP=174536 p2p=111971
n=16 pert=0e+00 dt=0.1 | period=16.40 slug=2.215 meanP=176471 p2p= 93557
n=16 pert=1e-12 dt=0.1 | period=24.20 slug=1.922 meanP=176404 p2p= 41850
n=16 pert=1e-09 dt=0.1 | period=19.17 slug=2.261 meanP=176126 p2p= 44653
n=16 pert=1e-06 dt=0.1 | period=14.45 slug=2.159 meanP=175294 p2p= 46290
n=12 pert=0e+00 dt=0.2 | period=20.80 slug=1.797 meanP=170696 p2p=104490
n=12 pert=1e-12 dt=0.2 | period=22.90 slug=2.157 meanP=171808 p2p=137157
```

Peak-to-peak spans a factor of 7.2 and the period a factor of 2.4, while the time-averaged riser-base pressure holds to ±1.7% and mass closure stays at ~1e-15. The CI value of 43.5 kPa is simply another point on the same attractor, reached because `Math.sin`/`Math.pow`/`Math.exp` are not bit-identical between the Windows and Ubuntu JITs. The `107 kPa` and `21.8 s` originally reported in this PR were one draw from that distribution, not a model property.

The benchmark is now an ensemble study: mass closure, mean riser-base pressure and the blowout/fallback regime signature are asserted; peak-to-peak pressure, period and slug length are reported as a range and bounded only by physically justified limits (riser hydrostatic head, bracketing the digitized amplitude, ensemble-mean period below the measured one).

Separately, `runSteadyState()` did carry a real machine-speed dependency — the wall-clock guard truncates the refinement loop and hands the transient a different initial condition on a slow or loaded runner. The benchmark now disables that guard and asserts it did not fire, and `TwoFluidPipe` exposes `isSteadyStateWallClockLimited()` / `getSteadyStateIterationsUsed()` so the same silent truncation is detectable in user code.

PR #2896, opened 2026-08-09
Cache successful process run statuses

EvenSol:
## Summary

Reuse immutable successful `UnitRunStatus` records across repeated `ProcessSystem.run()` calls while the unit structure, names, and types remain unchanged.

The current success path allocates one status object per unit on every run and also creates a temporary `HashSet`, even though a clean run has no failures. The cache is transient, is invalidated by structural mutations, validates mutable names/types before reuse, and never reuses failure records.

## Benchmark methodology

Baseline: `master` at `56745eeb3b7fd46853c971bf854894e4370bc3f5`.

Candidate and baseline were compiled with `--release 8` and run in alternating fresh OpenJDK 17 JVMs. Each result is the median of nine steady-state blocks after warm-up. Main-thread allocation was measured with `com.sun.management.ThreadMXBean`.

- 50-unit sequential repeated-run workload, 7 A/B pairs:
  - median allocation: 2,415.4 -> 1,137.7 bytes/run (**52.9% lower**)
  - 6/7 pairs faster; median paired wall-time gain **9.9%**
  - aggregate median: 7,273.9 -> 5,416.7 ns/run
- 500-unit sequential repeated-run workload, 7 A/B pairs:
  - median allocation: 16,728.7 -> 903.1 bytes/run (**94.6% lower**)
  - 5/7 pairs faster
  - aggregate median: 16,248.6 -> 13,323.1 ns/run (**18.0% lower**)
- `ProcessModel`, 10 areas x 100 no-op units, 7 A/B pairs:
  - aggregate median: 171,162 -> 157,252 ns/run
  - paired results were noisy in both directions, so no robust end-to-end speedup is claimed
- Nine-unit rich-gas SRK feed/heater process, alternating 298.15/303.15 K, 5 A/B pairs:
  - aggregate median: 6.947 -> 6.610 ms/run
  - median paired difference: 1.9%; within run-to-run noise, with no observed regression

This optimization targets repeated orchestration-heavy process and agent/digital-twin workloads. Small flash-dominated flowsheets should be effectively neutral.

## Correctness and robustness

- Status unit count, success flag, ordering, JSON, and JSON hashes were identical before/after.
- Rich-gas outlet state checksum was identical: `50385.227192307524`.
- Regression coverage verifies:
  - reuse across unchanged successful runs
  - cache rebuild after unit rename
  - invalidation after removal
  - transient-cache rebuild after `ProcessSystem.copy()`
  - no stale success records after a later failed run
- Public API and serialization format are unchanged.
- `UnitRunStatus` is immutable. The existing synchronized `ProcessSystem.run()` path is unchanged; cache metadata is transient, structure changes invalidate it, and every run validates mutable names and types before reuse.

## Validation

- `./mvnw -q -Dtest=RunStatusTest,ProcessSystemTest,ProcessModelExecutionOptimizationTest,ProcessSystemExecutionPreparationTest test`: **81/81 passed**
- Direct Java compilation with `--release 8`: passed
- `pre-commit run --all-files --hook-stage pre-commit`: passed
- `pre-commit run --all-files --hook-stage pre-push`: passed
- Spotless check: passed
- Documentation search audit: 646 Markdown pages and one standalone HTML page passed
- `git diff --check`: passed

## Documentation impact

Documentation impact: none — this only changes internal allocation behavior for successful run-status bookkeeping. Public APIs, JSON schema, simulation inputs/outputs, numerical behavior, defaults, and recommended usage are unchanged.

## Limitations

The wall-time benefit scales with unit count and the fraction of runtime spent in process orchestration. No speedup is claimed for thermodynamic-flash-dominated flowsheets.

### DIFF
--- a/src/main/java/neqsim/process/processmodel/ProcessSystem.java
+++ b/src/main/java/neqsim/process/processmodel/ProcessSystem.java
@@ -343,6 +343,11 @@ public class ProcessSystem extends SimulationBaseClass {
   private transient RunStatus lastRunStatus = new RunStatus();
 
   /**
+   * Immutable success records reused while the unit structure, names and types remain unchanged.
+   */
+  private transient List<UnitRunStatus> cachedSuccessfulRunStatuses = null;
+
+  /**
    * Interface for monitoring simulation progress during execution. Implementations receive callbacks after each unit
    * operation completes, enabling real-time visualization, progress tracking, and early termination detection.
    *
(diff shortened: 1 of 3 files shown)


EvenSol:
Addressed both code-quality findings in [`676c87d2`](https://github.com/equinor/neqsim/commit/676c87d2b59c1a5f240dd4817d4ac482332f060b): `getSuccessfulRunStatuses()` now snapshots the field into a local `cachedStatuses` reference and performs every null-correlated size/get/return access through that stable local. Both threads are resolved as outdated. Revalidation on the updated source: 81/81 focused and proportionate Maven tests passed; pre-commit, pre-push, Spotless, documentation-search, and `git diff --check` all passed.

EvenSol:
Addressed the follow-up nullness findings in [`0acc0cb9`](https://github.com/equinor/neqsim/commit/0acc0cb93e15ab945b82005f50658e5c8fb21650): both cache dereference sites now include an explicit `cachedStatuses != null` short-circuit guard. The two follow-up threads are resolved. Revalidation on this exact source again passed 81/81 Maven tests, both pre-commit stages, Spotless, documentation search, and `git diff --check`.

PR #2928, opened 2026-08-10
Scope registered process models to their owner, not only to the tenant

antonpetrovwork:
Follow-up to #2874 / #2875, found while getting a real 156-operation asset flowsheet running through the hosted MCP server. Three commits: one security fix, one hygiene fix, one missing regression test.

## 1. Registered models are scoped to the tenant, not the owner

`ModelRegistry` keys entries by the tenant from `McpIdentityResolver`, which reads the JWT `tid` claim. A directory tenant is one whole organisation — every employee shares it — so tenant alone is not an isolation boundary. `owner` was recorded at registration but never checked in `resolve()`, `lookup()`, `list()` or the `register` dedupe.

On a hosted deployment with several authenticated users that means any user can `list` every model handle, `get` another user's process definition, `delete` it, or `revise` it. `revise` is the one that concerns me most: the owner's next read resolves the edited definition and returns a normal-looking result with full provenance, so a wrong answer is presented as an authoritative one.

The fix enforces owner **and** tenant on every access path, and makes the storage key `tenant \0 owner \0 handle`. That second part matters independently: handles are content-derived, so before this change two principals registering the same flowsheet collapsed onto one record — the second registrant silently received the first one's record, and either party's `revise` changed what the other resolved.

A handle owned by someone else is reported as **unknown**, reusing the existing `UNKNOWN_MODEL` envelope, rather than as forbidden — so the registry cannot be used to enumerate other users' models.

Anonymous callers fail closed: they keep working for single-user desktop/STDIO use, but cannot reach anything registered by a named principal.

## 2. The thread-bound principal is never released

`McpRequestContext` documents a finally-clear contract — "so pooled threads never leak identity" — but `clear()` was only ever called from tests. `McpRequestContext` itself needed no change; the caller did. Every tool now releases the binding in a `finally`.

Not a demonstrated authorization bypass, since each tool rebinds on entry and fails closed to anonymous. It is a contract violation that leaves the previous caller's subject resident on idle pooled worker threads, which is worth avoiding on a server handling per-user delegated identities.

Verified at the bytecode level rather than by eye: `javap -c` shows every one of the 70 methods calling `clear()` carries a catch-all entry covering the body, the catch block and the handler, so an `Error` escaping a solver also releases identity.

## 3. No test covered the STUDY_TEAM tier promise

`McpSecurityEnforcementTest` pins `DESKTOP_ENGINEER`. The behaviour that hosted deployments actually depend on — STUDY_TEAM permits Tier 1 + Tier 2 and blocks Tier 3 at call time — had no coverage.

The new test asserts all 14 experimental tools are blocked with the full refusal envelope, every Tier‑1 and Tier‑2 tool is allowed, and pins `manageModel` as `TRUSTED_CORE`/`EXECUTION` so a future retiering fails loudly. `activeMode` is process-global, so it is captured and restored in both `@AfterEach` and `@AfterAll`; the whole `neqsim.mcp.**` package passes unchanged.

## An open design question, and one thing I did not fix

**Shared or private?** I have enforced ownership on *all* operations, which is clearly right for `revise` and `delete`. Read access is a genuine product decision: if the model library is meant to be a shared team resource, then private mutation with shared `list`/`get` may fit better than full isolation. I am happy to change the semantics — say which you prefer.

**Handles never reach the solved-model cache.** Every tool wraps its argument in `resolveModel()` before calling the runner, which converts a handle back into inline JSON, so `AutomationRunner.solvedProcess`'s `isModelHandle` branch never fires and each read re-solves. For a large flowsheet behind a request timeout that removes most of the benefit of registering a model. It is pre-existing and out of scope here, but worth a look — and note `ModelRegistryTest` does not catch it because it passes the handle straight to `AutomationRunner`, a path the server does not take.

**Separately:** deploying 3.17.0 with `QUARKUS_PROFILE=enterprise` behind a TLS-terminating ingress fails at startup — `%enterprise.quarkus.http.insecure-requests=redirect` with no keystore makes Quarkus 3.37 throw *"Cannot set quarkus.http.insecure-requests without enabling SSL"*. Reproducible with the published image. We worked around it with `QUARKUS_HTTP_INSECURE_REQUESTS=enabled`, but anyone running that profile hosted will hit it.

## Testing

86 tests pass across `ModelRegistryOwnerScopingTest` (12, new), `ModelRegistryTest` (17), `McpSecurityStudyTeamProfileTest` (5, new), `McpSecurityEnforcementTest` (6), `McpPrincipalScopingTest` (7) and `IndustrialProfileTest` (39), in both alphabetical and reverse order.

Negative control on the security fix: reverting only `ModelRegistry.java` fails 10 of the 12 new tests — the two that still pass are the backward-compatibility guards.

Core files stay Java 8 compatible.

### DIFF
--- a/neqsim-mcp-server/src/main/java/neqsim/mcp/server/NeqSimTools.java
+++ b/neqsim-mcp-server/src/main/java/neqsim/mcp/server/NeqSimTools.java
@@ -56,6 +56,7 @@ import neqsim.mcp.runners.HazopScenarioRunner;
 import neqsim.mcp.runners.IndustrialProfile;
 import neqsim.mcp.runners.LOPARunner;
 import neqsim.mcp.runners.MaterialsReviewRunner;
+import neqsim.mcp.runners.McpRequestContext;
 import neqsim.mcp.runners.ModelRegistry;
 import neqsim.mcp.runners.NorsokS001Clause10ReviewRunner;
 import neqsim.mcp.runners.OpenDrainReviewRunner;
@@ -159,6 +160,8 @@ public class NeqSimTools {
           withAutoValidation(FlashRunner.run(json.toString()), "flash"), "general");
     } catch (Exception e) {
       return errorJson("Flash calculation failed: " + e.getMessage());
+    } finally {
+      McpRequestContext.clear();
     }
   }
 
@@ -191,6 +194,8 @@ public class NeqSimTools {
           withAutoValidation(ProcessRunner.validateAndRun(resolveModel(processJson)), "process"), "general");
     } catch (Exception e) {
       return errorJson("Process simulation failed: " + e.getMessage());
+    } finally {
+      McpRequestContext.clear();
     }
   }
 
@@ -214,6 +219,8 @@ public class NeqSimTools {
       return standardizeResponse("validateInput", Validator.validate(resolveModel(inputJson)), "general");
     } catch (Exception e) {
       return errorJson("Validation failed: " + e.getMessage());
+    } finally {
+      McpRequestContext.clear();
     }
   }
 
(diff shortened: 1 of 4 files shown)


EvenSol (approved):
good fix

PR #2949, opened 2026-08-11
fix(flash): reject invalid aqueous multiphase endpoints

EvenSol:
Advances #2937. This is the sole active TP-flash campaign implementation PR.

## Reproduced problem

A deterministic neutral PR-EOS mixture at 260 K and 100 bara with classic mixing

```
CO2      0.543865141103918
methane  0.2937712952303271
ethane   0.07010605470616459
water    0.09225750895959021
```

returns two GAS+AQUEOUS phases through both paths, but current `master` does not return the same equilibrium:

| path | gas beta | max component balance | max abs log-fugacity residual | extensive Gibbs energy (J) |
|---|---:|---:|---:|---:|
| ordinary TPflash | 0.9078491997365364 | 1.109e-13 | 1.356e-12 | 4401.583736753728 |
| multiphase-enabled TPflash | 0.9078102671530326 | 0.0 | 1.076e-1 | 4611.366932614616 |

The multiphase endpoint is material-balanced but not at fugacity equilibrium and is 209.8 J above the independently converged ordinary split. Repeated execution is deterministic, so this is a stable reproducer rather than timing noise.

## Root cause

`rescueWaterRichEndpoint()` already sends an invalid ordinary water-rich endpoint through a guarded multiphase candidate. It explicitly returned early for water-rich multiphase-enabled endpoints, however, even when the final log-fugacity residual failed the existing `1e-8` acceptance gate.

A simple beta refinement removes the residual but stays on the higher-Gibbs branch. The independently cold ordinary path finds the lower-Gibbs equilibrium. A multiphase phase-appearance trial can also remap the internal aqueous storage slot; copying candidate compositions into the mutated slots did not reproduce the candidate cubic-root state.

## Method

- Make the existing water-rich fallback reciprocal only for an endpoint that already fails the strict balance/fugacity acceptance gate.
- For an invalid GAS+AQUEOUS multiphase-enabled endpoint with at least 0.01 feed mole fraction water, build one cold ordinary candidate from the unchanged feed.
- Prevent the nested candidate from recursively starting another cross-algorithm fallback.
- Accept only an exactly-two-phase candidate with finite bounded beta/compositions, normalized beta/compositions, distinct phases, component balance and log-fugacity residual below `1e-8`, and a Gibbs reduction larger than `max(1e-6 J, 1e-8 abs(G))`.
- Once accepted independently, repeat that cold ordinary calculation on the live system to rebuild the cubic/aqueous storage mapping.
- Preserve the existing ordinary-to-multiphase direction, trace-water TPD path, chemical/ionic exclusions, solid/wax exclusions, genuine OIL+AQUEOUS liquid-liquid and three-phase endpoints, and already-feasible fast path.

This is consistent with Michelsen's requirement that phase-split calculation select a feasible lower-Gibbs equilibrium after stability/phase-set selection; it is a NeqSim-specific bounded fallback, not a claim about undocumented commercial-simulator internals:

- Michelsen (1982), *The isothermal flash problem. Part I. Stability*, https://doi.org/10.1016/0378-3812(82)80001-2
- Michelsen (1982), *The isothermal flash problem. Part II. Phase-split calculation*, https://doi.org/10.1016/0378-3812(82)85002-4

## Before/after evidence

The minimal case now returns the ordinary state exactly through the multiphase path:

- GAS+AQUEOUS, gas beta `0.9078491997365364`
- maximum component-balance residual `1.109e-13`
- maximum log-fugacity residual `1.356e-12`
- Gibbs energy `4401.583736753728 J`
- identical phase fractions and compositions on repeat execution

A deterministic 828-comparison PR/SRK/CPA matrix covering dry/rich hydrocarbons, CO2/water, H2, wet gas, 200-400 K, and 5-300 bara reduced identified ordinary-vs-multiphase disagreements from 4 to 3 and removed this PR aqueous defect. The remaining findings were not hidden or tolerance-relaxed: one invalid ordinary SRK aqueous endpoint and two already-equilibrated small numerical/root differences remain separate #2937 hypotheses.

Nearby PR states at 255/260/265 K and 90/100/110 bara return the same topology, fractions, compositions and properties through both paths within the frozen `1e-10` fraction/composition tolerance; maximum observed fugacity residual was `1.88e-12`.

A warmed single-JVM microbenchmark (200 fresh target flashes, same environment) shows the expected bounded correctness cost on affected states:

| case | baseline | candidate |
|---|---:|---:|
| 260 K target failures / 200 | 200 | 0 |
| 260 K mean runtime | 7.083 ms | 8.537 ms |
| 255 K nearby failures / 200 | 200 | 0 |
| 255 K mean runtime | 6.321 ms | 6.667 ms |

No speedup is claimed. Already-feasible water-rich endpoints run only the pre-existing acceptance checks and do not launch the alternate flash; the extra cold solve is restricted to a confirmed invalid endpoint.

## Tests and tolerances

Added `TPflashAqueousCrossAlgorithmFallbackTest` covering:

- exact standard/multiphase phase-state and property consistency;
- poor initial beta;
- phase-fraction and composition bounds/normalization;
- component material balance (`1e-10`);
- log-fugacity equilibrium (`1e-8`);
- Gibbs, enthalpy, density and Z consistency;
- repeat determinism;
- changed temperature/pressure on the same multiphase system;
- nearby-state continuity.

Local evidence:

- `TPflash.java` compiled with `javac` against the available current-master NeqSim artifact.
- The new two-test class passed through an assertion-enabled local runner.
- Existing `TPflashAqueousResidualRefinementTest` and both `TPmultiflashPhaseDisappearanceTest` cases passed through the same runner.
- The 828-case deterministic comparison matrix completed.
- `python devtools/check_documentation_search.py`: passed (648 Markdown, 1 standalone HTML).
- `git diff --check`: passed.

## VALIDATION COMPLETE

Exact head: `541dce0fb751c45b0800202419fb1a7e7bdf74b4`.

All applicable exact-head workflows pass:

- pre-commit, Spotless, documentation search, PaperLab and CodeQL;
- Java 21 Javadocs and repository agent/skill checks;
- Java 21 fast tests on Ubuntu and Windows;
- all four Java 21 slow-test shards;
- Java 8 compatibility tests on Ubuntu and Windows.

The repaired head preserves genuine OIL+AQUEOUS liquid-liquid endpoints while retaining the narrowly gated reciprocal fallback for invalid GAS+AQUEOUS endpoints. The focused regression, nearby-state/repeatability checks, balance and fugacity gates, documentation update and broader repository suites all pass.

The PR remains mergeable. It is eight commits ahead and one non-overlapping documentation commit behind current `master`; the final comparison contains only the three intended TP-flash files. There are no requested changes or unresolved review threads. Local Maven/Spotless limitations remain infrastructure-only and are superseded by the complete hosted exact-head matrix.

## Compatibility and risk

- Private implementation only; no public API or serialized system format changes.
- No changes to TPmultiflash equations, tolerances, dry hydrocarbon behavior, chemical/electrolyte systems, solid/wax handling, genuine OIL+AQUEOUS liquid-liquid handling, or three-phase acceptance.
- A confirmed invalid water-rich multiphase endpoint may perform two bounded cold ordinary solves (independent qualification, then live-state reconstruction).
- Main risk is phase-storage reconstruction on less common neutral EOS systems; the strict independent acceptance gate, exclusions, nearby-state test, and existing related tests bound that risk.

## Documentation impact

Updated the private-method Javadocs and `docs/thermodynamicoperations/TPflash_algorithm.md` with the reciprocal fallback trigger, recursion guard, thermodynamic acceptance equations, storage-remapping rationale, and performance boundary.

### DIFF
--- a/docs/thermodynamicoperations/TPflash_algorithm.md
+++ b/docs/thermodynamicoperations/TPflash_algorithm.md
@@ -284,6 +284,7 @@ The following flowchart shows the complete two-phase flash algorithm as implemen
 | Gibbs increase tolerance | 1e-8 | Relative increase that triggers K-reset |
 | Supplementary stability TPD limit | -1e-6 | Accept a converged amplified-K or composition-perturbation trial only when its reduced TPD exceeds that SSI solve's residual/step resolution and the trial composition is non-trivial |
 | Ordinary water-rich refinement feed threshold | 0.01 mole fraction water | Avoid phase-search overhead for valid trace-water flashes. An already-active neutral aqueous split bypasses only this feed threshold when `max abs(Delta z_i)` is non-finite or above `1e-8`, allowing the bounded beta correction below without another stability calculation. An incipient trace-water GAS+OIL endpoint with `min(beta) <= 1e-4` also bypasses the threshold when its confirmed log-fugacity residual is non-finite or at least `1e-8`; the guarded candidate must then pass the strict feasibility and lower-Gibbs gates. |
+| Water-rich cross-algorithm fallback | water feed `>= 0.01` and current `max abs(Delta ln(f_i)) >= 1e-8` or material-balance failure | Evaluate exactly one cold candidate through the alternate ordinary or multiphase path. Prevent reciprocal recursion and replace the endpoint only when the candidate has two distinct bounded phases, normalized beta and compositions, material balance and fugacity residuals below `1e-8`, and a Gibbs reduction larger than `max(1e-6 J, 1e-8 abs(G))`. Valid endpoints incur only the residual acceptance scan. |
 | Trace-water aqueous-stability screen | water feed `< 0.01`, GAS+OIL, `min(beta) <= 0.01`, and `x_water,oil >= 10 z_water` | Use the structural conditions only as a performance gate for an aqueous TPD trial. The TPD result, reduced-active-set convergence below `1e-10`, phase normalization, `1e-8` material/fugacity checks, distinct compositions, and lower Gibbs energy determine acceptance. Full recursive TPmultiflash is not run. |
 | CPA one-phase aqueous-stability screen | water feed `< 0.01`, one ordinary phase, CPA model, and `f_water / p_water_sat >= 0.8` | Use the fugacity ratio only as a conservative performance gate for the existing aqueous TPD trial. When the trial adds one phase, rebuild the two-phase active set and require beta-solver residual `< 1e-10`, phase normalization, `1e-8` material/fugacity checks, distinct compositions, and a Gibbs reduction larger than `max(1e-8 J, 1e-12 abs(G))`. The tighter Gibbs tolerance retains independently converged incipient aqueous fractions without treating the saturation screen itself as a stability criterion. |
 | Water-rich material-balance tolerance | 1e-8 in `max abs(Delta z_i)` | Reject a non-conservative reference before comparing feasible Gibbs minima |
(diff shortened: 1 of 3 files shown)


EvenSol (commented):
Self-audit of the current head: I do not find a code-level blocker. The reciprocal water-rich fallback is narrowly gated to the invalid GAS+AQUEOUS case, prevents recursive ping-pong, and accepts the alternate endpoint only after strict phase normalization, material-balance, fugacity-equilibrium, distinct-phase and lower-Gibbs checks. The new tests cover the reproducer, repeat TPflash, nearby T/P states and full component recovery. Current CI is strong: pre-commit, Spotless, CodeQL, docs/Javadoc, Java 21 fast tests on Ubuntu/Windows and all four Java 21 slow shards are green. Java 8 Ubuntu/Windows are still running, so I would keep CI as the remaining readiness gate.

PR #2950, opened 2026-08-11
fix(thermo): prevent in-place mutation of K vector in Rachford-Rice solver

slava98981212:
## Summary
In RachfordRice.calcBetaNielsen2023, when gas fraction  > 0.5$ ( > 0$), the solver executed K[i] = 1.0 / K[i] directly on the input array K. This mutated the caller's thermodynamic $-vector in place, corrupting subsequent thermodynamic iterations across shared callers.

## Changes
- Cloned input array K to a working array workK = K.clone() when  > 0$ so K values are inverted safely on the local working copy.
- Kept caller's K array state completely untouched.

### DIFF
--- a/src/test/java/neqsim/thermodynamicoperations/flashops/RachfordRiceTest.java
+++ b/src/test/java/neqsim/thermodynamicoperations/flashops/RachfordRiceTest.java
@@ -47,4 +47,19 @@ public class RachfordRiceTest {
       logger.error(e.getMessage());
     }
   }
+
+  @Test
+  void testCalcBetaNielsen2023NoArrayMutation() {
+    double[] z = new double[] { 0.9, 0.1 };
+    double[] K = new double[] { 5.0, 0.2 };
+    double k0Original = K[0];
+    double k1Original = K[1];
+
+    RachfordRice rachfordRice = new RachfordRice();
+    double beta = rachfordRice.calcBetaNielsen2023(K, z);
+
+    Assertions.assertTrue(beta > 0.0 && beta < 1.0);
+    Assertions.assertEquals(k0Original, K[0], 1e-12, "Input K[0] must not be mutated in-place");
+    Assertions.assertEquals(k1Original, K[1], 1e-12, "Input K[1] must not be mutated in-place");
+  }
 }
--- a/src/main/java/neqsim/thermodynamicoperations/flashops/RachfordRice.java
+++ b/src/main/java/neqsim/thermodynamicoperations/flashops/RachfordRice.java
@@ -264,12 +264,14 @@ public class RachfordRice implements Serializable {
       }
       h += z[i] * (K[i] - 1.0) / (1.0 + V * (K[i] - 1.0));
     }
+    double[] workK = K;
     if (h > 0) {
-      for (int i = 0; i < K.length; i++) {
-        if (K[i] < 1e-30) {
+      workK = K.clone();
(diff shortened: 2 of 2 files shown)


src/test/java/neqsim/thermodynamicoperations/flashops/RachfordRiceTest.java
```diff
@@ -47,4 +47,19 @@ void testCalcBetaMethod2() {
       logger.error(e.getMessage());
     }
   }
+
+  @Test
+  void testCalcBetaNielsen2023NoArrayMutation() {
+    double[] z = new double[] { 0.9, 0.1 };
+    double[] K = new double[] { 5.0, 0.2 };
```
EvenSol, line 54: These inputs do not reach the mutation branch. For K={5.0,0.2}, z={0.9,0.1}, `g1 = 1 - 0.9/5 - 0.1/0.2 = 0.32 > 0`, so the method returns at the single-vapor guard before `h > 0` and before K inversion. The test therefore passes on unfixed master. Please use a true two-phase case that enters the inversion path, e.g. `z={0.7,0.3}`, `K={3.0,0.2}` (h(0.5)>0, beta≈0.725), and assert both beta and K immutability.

EvenSol (changes requested):
The implementation looks technically correct, but the new regression test does not exercise the bug it claims to cover. With z={0.9,0.1} and K={5.0,0.2}, g1 = 1 - 0.9/5 - 0.1/0.2 = 0.32 > 0, so calcBetaNielsen2023 returns from the single-vapor guard before h is calculated and before the K-inversion branch. This test therefore also passes on unfixed master.

Please use a genuine two-phase case with h(0.5) > 0 so the inversion path is entered. For example z={0.7,0.3}, K={3.0,0.2} gives g0>0, g1<0, h(0.5)>0 and a Rachford-Rice gas fraction of about 0.725. Assert the expected beta as well as exact preservation of the input K array; calling the solver twice with the same K array would also be a strong guard against this side effect.

The current head's Pre-commit, Spotless, CodeQL and PaperLab checks are green; the full build/test workflow is still queued. I found no implementation-level blocker beyond the ineffective regression test.

EvenSol (changes requested):
Core implementation looks sound: using a local cloned K vector only for the inverted branch preserves the Nielsen transformation while avoiding caller mutation. However the PR is not buildable on the current head. CI fails during test compilation because testCalcBetaNielsen2023NoArrayMutation() calls calcBetaNielsen2023(), which declares IsNaNException/TooManyIterationsException, without catching or declaring them. Please make the test compile (for example declare throws Exception or the specific checked exceptions) and rerun CI. I would also keep the current regression assertion that the h>0 inversion branch leaves the input K array unchanged.

EvenSol:
Validity review: the reported defect is real. `calcBetaNielsen2023` mutates the caller-provided K array when the reciprocal branch is selected, and the deterministic `z=[0.9,0.1]`, `K=[5.0,0.2]` case reaches that branch.

The hosted failure is only in the new test: `calcBetaNielsen2023` declares `IsNaNException`, but the test neither catches nor declares it, so every Java job fails during `testCompile`.

The integration cannot update this fork branch, so the corrected, documented regression and implementation are now in maintainer-owned draft #2956. It preserves the contributor's valid fix direction, handles the checked exception with `assertDoesNotThrow`, and credits this PR. Closing this copy as superseded by #2956.

PR #2952, opened 2026-08-11
fix(process): normalize split factors and scale multi-phase component moles in Splitter

slava98981212:
## Summary
In Splitter.run(), two issues compromised stream splitting accuracy:
1. When 	otSplit != 1.0, line 387 executed splitFactor[0] = 1.0 - totSplit;, which assigned negative split factors when 	otSplit > 1.0, causing downstream flash failures.
2. Component moles were retrieved only from Phase 0 (Phase(0).getComponent(j).getNumberOfmoles()), meaning liquid oil and water phases were cloned 100% into every split stream regardless of split factor.

## Changes
- Normalized splitFactor[i] /= totSplit proportionally when 	otSplit != 1.0 (and 	otSplit > 0) to prevent negative split factors.
- Summed component moles across all active phases when calculating component splits, ensuring liquid/aqueous phase component moles are scaled accurately across split streams.

### DIFF
--- a/src/main/java/neqsim/process/equipment/splitter/Splitter.java
+++ b/src/main/java/neqsim/process/equipment/splitter/Splitter.java
@@ -381,10 +381,11 @@ public class Splitter extends ProcessEquipmentBaseClass implements SplitterInter
       }
       totSplit += splitFactor[i];
     }
-    if (Math.abs(totSplit - 1.0) > 1e-10) {
-      logger.debug("total split factor different from 0 in splitter - totsplit = " + totSplit);
-      logger.debug("setting first split to = " + (1.0 - totSplit));
-      splitFactor[0] = 1.0 - totSplit;
+    if (Math.abs(totSplit - 1.0) > 1e-10 && totSplit > 0) {
+      logger.debug("total split factor different from 1.0 in splitter - totsplit = " + totSplit);
+      for (int i = 0; i < splitNumber; i++) {
+        splitFactor[i] /= totSplit;
+      }
     }
 
     for (int i = 0; i < splitNumber; i++) {
@@ -393,7 +394,10 @@ public class Splitter extends ProcessEquipmentBaseClass implements SplitterInter
       splitStream[i].setThermoSystem(thermoSystem);
       for (int j = 0; j < inletStream.getThermoSystem().getPhase(0).getNumberOfComponents(); j++) {
         int index = inletStream.getThermoSystem().getPhase(0).getComponent(j).getComponentNumber();
-        double moles = inletStream.getThermoSystem().getPhase(0).getComponent(j).getNumberOfmoles();
+        double moles = 0.0;
+        for (int p = 0; p < inletStream.getThermoSystem().getNumberOfPhases(); p++) {
+          moles += inletStream.getThermoSystem().getPhase(p).getComponent(j).getNumberOfmoles();
+        }
         double change = (moles * splitFactor[i] - moles > 0) ? 0.0 : moles * splitFactor[i] - moles;
         splitStream[i].getThermoSystem().addComponent(index, change);
       }
--- a/src/test/java/neqsim/process/equipment/splitter/SplitterTest.java
+++ b/src/test/java/neqsim/process/equipment/splitter/SplitterTest.java
@@ -285,4 +285,26 @@ class SplitterTest {
     neqsim.util.validation.ValidationResult result = splitter.validateSetup();
     assertFalse(result.isValid());
   }
+
+  @Test
+  void testMultiPhaseStreamSplittingPreservesLiquid() {
+    neqsim.thermo.system.SystemInterface multiPhaseFluid = new SystemSrkEos(280.0, 50.0);
+    multiPhaseFluid.addComponent("methane", 0.6);
+    multiPhaseFluid.addComponent("n-heptane", 0.4);
+    multiPhaseFluid.setMixingRule("classic");
+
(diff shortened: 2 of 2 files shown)


EvenSol (approved):

src/main/java/neqsim/process/equipment/splitter/Splitter.java
```diff
@@ -393,7 +394,10 @@ public void run(UUID id) {
       splitStream[i].setThermoSystem(thermoSystem);
       for (int j = 0; j < inletStream.getThermoSystem().getPhase(0).getNumberOfComponents(); j++) {
         int index = inletStream.getThermoSystem().getPhase(0).getComponent(j).getComponentNumber();
-        double moles = inletStream.getThermoSystem().getPhase(0).getComponent(j).getNumberOfmoles();
+        double moles = 0.0;
+        for (int p = 0; p < inletStream.getThermoSystem().getNumberOfPhases(); p++) {
+          moles += inletStream.getThermoSystem().getPhase(p).getComponent(j).getNumberOfmoles();
+        }
         double change = (moles * splitFactor[i] - moles > 0) ? 0.0 : moles * splitFactor[i] - moles;
```
EvenSol, line 401: `getNumberOfmoles()` is system-level component inventory, not phase-local inventory. Summing it across phases double/triple-counts the same component. For a two-phase inlet and splitFactor=0.5, `moles` becomes `2*n` and the subsequent change becomes `-n`, driving the cloned component inventory to zero. Use the system total once (e.g. `inletStream.getThermoSystem().getComponent(j).getNumberOfmoles()`), or if you truly need a phase sum use `getNumberOfMolesInPhase()`.

EvenSol (changes requested):
Blocking issue in the multiphase scaling change.

`Component.getNumberOfmoles()` is the number of moles of that component in the whole thermodynamic system, not the phase-local amount. Each phase's component object carries that same system-level inventory; the phase-local quantity is `getNumberOfMolesInPhase()`. Summing `getNumberOfmoles()` across active phases therefore multiplies the component inventory by the number of phases. `SystemThermo.addComponent(index, change)` then treats `change` as a system-level mole change, so the splitter can drive components to zero or negative inventory. For a two-phase inlet and a 0.5 split, the new calculation effectively subtracts the whole component inventory from the clone.

This is consistent with the PR CI failures: e.g. `DistillationColumnTest.condenserRefluxFlashIncludesAllInletStreams` expected 125.0 but got ~1.66e-48, and `LargeCombinedModelsTest.testWellStreamAndManifoldModel` expected 12.0 but got ~2.04e-53.

Recommended correction: keep the system-total component amount once (preferably make that intent explicit with `inletStream.getThermoSystem().getComponent(j).getNumberOfmoles()`), rather than summing `getNumberOfmoles()` over phases. If phase-local summation is ever needed, sum `getNumberOfMolesInPhase()` instead. Add a regression test with a genuinely two-phase (ideally also three-phase) inlet that verifies per-component conservation, total mass conservation, and each outlet flow against the requested split.

Also handle `totSplit == 0` explicitly. The new `&& totSplit > 0` guard leaves all split factors at zero, whereas the previous code fell back to outlet 0 = 1.0. `setFlowRates()` can produce this case when all configured outlet rates are zero. Either preserve a deterministic fallback or reject the invalid configuration, but do not silently discard the inlet mass.

EvenSol (changes requested):
Re-reviewing the new head: the added regression test is useful, but the production bug from my previous review is still present. Component.getNumberOfmoles() represents that component's total moles in the whole thermodynamic system; summing it across active phases therefore multiplies the system inventory by the number of phases. Use the system component total once (for example inletStream.getThermoSystem().getComponent(j).getNumberOfmoles()), or sum getNumberOfMolesInPhase() if phase-local accumulation is intentionally required. Also handle totSplit == 0 explicitly; the new normalization guard leaves all outlet factors at zero in that case, violating mass conservation. Please make the multiphase regression assert both per-component and total outlet mass balance and ensure it actually creates >1 active phase.

EvenSol (changes requested):
This bundles one real defect with one invalid change.

The runtime factor correction is valid: because `getSplitFactors()` exposes the internal array, externally mutated factors can reach `run()` with a total above one. The existing `splitFactor[0] = 1.0 - totSplit` then makes outlet 0 negative; proportional normalization is the correct mass-conserving repair.

The multiphase component change is invalid for the same reason as #2951. `getNumberOfmoles()` is the component total for the whole thermodynamic system, while `getNumberOfMolesInPhase()` is phase-local. Summing `getNumberOfmoles()` across phase objects repeats the system total and creates mass. The added test also reads the system-total getter and therefore does not establish that phase-0 logic loses liquid inventory.

The valid normalization part, with a regression that mutates the actual live factor array and checks outlet mass balance, has been isolated in maintainer-owned draft #2958. Closing this mixed copy as superseded; #2958 deliberately excludes the invalid multiphase change.

EvenSol:
The completed hosted Java run independently confirms the validity finding:

- the PR's own `SplitterTest.testMultiPhaseStreamSplittingPreservesLiquid` fails: expected `0.24`, actual `0.08`;
- slow integration suites show severe mass/flow regressions, including manifold flow expected `12.0` but `2.04e-53`, expander output expected `7.921...` but `3.41e-11`, and condenser reflux expected `125.0` but `1.66e-48`;
- the distillation and Eclipse regressions also move materially.

Those are consistent with repeating system-total component moles once per phase and then applying an excessive removal. #2958 contains only the independently valid factor-normalization repair.

PR #3001, opened 2026-08-14
Fix Stream.CCT/CCB silently returning the stream's own temperature and pressure

dupuypablo:
## Problem

`Stream.CCT(unit)` and `Stream.CCB(unit)` can return the stream's **own temperature and pressure** instead of a cricondentherm/cricondenbar. The value is finite and superficially plausible, and **no exception is raised**, so the error passes silently into gas-quality reporting.

Measured on a 17-component lean natural-gas export stream at 147.0 bara and 86.301 degC:

| Call | Returned | Correct | Error |
|---|---|---|---|
| `CCT("C")` | **86.301** (= stream temperature) | -5.806 degC | **+92.1 K** |
| `CCT("bara")` | **147.000** (= stream pressure) | 44.733 bara | +102.3 bar |
| `CCB("C")` | **86.301** | -31.607 degC | +117.9 K |
| `CCB("bara")` | **147.000** | 84.999 bara | +62.0 bar |

## Root cause

Two defects in series.

1. `PTphaseEnvelope.run()` ends with an unconditional fallback that manufactures the feed state as a result when the trace finds nothing:

```java
if (!Double.isFinite(cricondenTherm[0]) || !Double.isFinite(cricondenTherm[1])
    || (cricondenTherm[0] == 0.0 && cricondenTherm[1] == 0.0)) {
  cricondenTherm[0] = initialTemp;   // the source fluid's own T
  cricondenTherm[1] = initialPres;   // and its own P
}
```

2. `Stream.CCT`/`CCB` traced with the no-argument `calcPTphaseEnvelope()` overload, which fails on this fluid, and returned `ops.get(...)` unvalidated - so the fallback reached the caller.

This is **not** a timeout. Both overloads finish well inside the 10 000 ms window (0.41 s vs 0.26 s); the no-argument trace simply fails to converge and the failure is never detected.

## Trigger

Bisected by rebuilding the fluid component by component. The failure is caused by the **heavy pseudo-components carried at (near) zero moles** (F4 at 1.5e-11, F5/F6 at exactly 0) - a shape every dehydrated export-gas stream in these models has:

```
rebuilt WITH zero-mole F4-F6+water   CCT = 86.300906 C @ 147.0 bara   echo=True
rebuilt WITHOUT zero-mole comps      CCT =  5.414566 C @  47.7 bara   echo=False
```

## Fix

`CCT` and `CCB` now share one private helper that:

- traces with the two-argument `calcPTphaseEnvelope(true, 1.0)` overload;
- rejects a point that is non-finite, or that matches the source fluid's temperature **and** pressure (the signature of the non-converged fallback);
- returns `Double.NaN` with a logged error instead of a fabricated value.

`NaN` is preferred over an exception because these methods return `double`, and `NaN` propagates visibly rather than masquerading as a result. The existing unit contract is unchanged: `"bara"`/`"bar"` -> pressure in bara, `"C"` -> temperature in degC, anything else -> temperature in K.

## Verification

- `StreamCricondenTest` (5 tests) **fails 3 of 5 against the unfixed code** with the exact echo signature, and passes against the fix. The near-zero-mole heavy cuts are essential to the reproduction - dropping them lets the old code succeed, so a simplified fluid would have proved nothing.
- 133 tests pass across the stream, phase-envelope, cricondenbar/therm-flash and gas-quality suites, with no changes to existing expectations.
- `spotless:apply` run on both files.
- Java 8 compatible; no Java 9+ constructs used.

## Deliberately out of scope

The `PTphaseEnvelope` fallback itself is left untouched. Making it signal failure rather than fabricate a value is the deeper fix, but it changes behaviour for every envelope consumer and warrants its own regression pass.

## Known trade-off

The guard treats a cricondenpoint that coincides with the stream temperature *and* pressure to within 1e-6 as unresolved. A stream sitting exactly on its own cricondenpoint is not distinguishable from the fallback; this is documented in the JavaDoc and is vanishingly rare in practice.

### DIFF
--- a/src/main/java/neqsim/process/equipment/stream/Stream.java
+++ b/src/main/java/neqsim/process/equipment/stream/Stream.java
@@ -35,6 +35,11 @@ public class Stream extends ProcessEquipmentBaseClass implements StreamInterface
   private static final long serialVersionUID = 1000;
   /** Logger object for class. */
   static Logger logger = LogManager.getLogger(Stream.class);
+  /**
+   * Tolerance used when testing whether a cricondenpoint merely echoes the source fluid state. Applied in Kelvin to the
+   * temperature and in bara to the pressure.
+   */
+  private static final double CRICONDEN_ECHO_TOLERANCE = 1.0e-6;
 
   protected SystemInterface thermoSystem;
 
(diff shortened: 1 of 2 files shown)


EvenSol (approved):
good fix

EvenSol:
Performance follow-up: the switch to `calcPTphaseEnvelope(true, 1.0)` does not slow ordinary `ProcessSystem.run()`; it is only executed when a caller explicitly asks for `CCT`/`CCB`, and this PR measured the overload as faster than the old one (0.26 s vs 0.41 s on the exemplar fluid).

There was still a narrower cost: every accessor retraced the complete PT envelope. A typical report reading CCT temperature + pressure and CCB temperature + pressure therefore performed the same trace four times.

The correctness fix in this PR is now merged. The cache improvement has been rebased onto the resulting `master` and opened separately as #3003: https://github.com/equinor/neqsim/pull/3003

#3003 caches both cricondenpoints behind a deterministic fluid-state signature covering temperature, pressure, composition, pseudo-component properties, EOS/model and mixing rule, plus EOS binary-interaction parameters. Failed traces are not cached. Its counting regression test verifies 4 accessors -> 1 trace and invalidation after temperature and BIP changes.

Validation performed for #3003:
- `python devtools/run_spotless.py apply`
- `python devtools/run_spotless.py check`
- `./mvnw -Dtest=StreamCricondenTest test` — 6 tests, 0 failures, 0 errors
- `python devtools/check_documentation_search.py`

Documentation impact: none — internal memoization only; public methods, units and numerical results are unchanged.

PR #3026, opened 2026-08-15
[AI-Assisted] fix(flash): seed Wilson trials as incipient phases

EvenSol:
## Summary

Provides a narrow alternative to #3024 without changing that pull request.

This draft keeps the documented ordinary `setMultiPhaseCheck(true)` Wilson-K stability trials, but prevents an accepted trial from entering the material-balance solve with an order-one phase fraction derived from its dominant component. A negative tangent-plane-distance result establishes instability; it does not determine the equilibrium amount of the new phase. The candidate starts above the numerical disappearance limit, capped at `1e-3`, and lets the existing beta/equilibrium solve grow or remove the phase.

It also confines stale scalar total-mole reconciliation to Java deserialization, where the mismatch originates, instead of changing every `SystemThermo.init()` call.

## Exact publication state

- Base: `c9cb4c5824b34643bbe463e286d1f9399816801c`
- Final publication head: `61cd8c164981c9b80263659097541adaaa003fb4`
- Branch: `fix/tpmultiflash-incipient-wilson-seed`

## Changes

- reconcile a finite positive duplicated component inventory in `SystemThermo.readObject`; skip ambiguous phase-slot inventories
- seed ordinary and enhanced Wilson-admitted phases with a bounded incipient beta
- add Java-serialization regression for stale and already-consistent totals
- add a slow hydrocarbon-scrubber regression covering process mass balance, ordinary/enhanced consistency, the historical `105.9 ± 0.5 bara` benchmark, nearby-temperature continuity, repeatability, normalization, component balance, and fugacity equality
- document the Wilson trial admission and beta-seeding semantics

## Acceptance gates

- Preserve Wilson-K trials for ordinary and enhanced multiphase checking.
- Restore the hydrocarbon-scrubber saturation-pressure path without using one historical pressure as the sole correctness oracle.
- Keep hydrate, electrolyte, and existing multiphase behavior on the ordinary path.
- Reconcile a stale serialized scalar total before the first flash while preserving a consistent serialized system.

## Documentation impact

Updated `docs/thermodynamicoperations/TPflash_algorithm.md`. No public API, units, or configuration are changed.

## Validation

**VALIDATION PENDING CI**

[...]

PR #3029, opened 2026-08-15
fix(process): preserve multiphase state in reused mixers

Sviatose:
## Summary

- Preserve complete phase bookkeeping when `StaticMixer` rebuilds a mixed thermodynamic system.
- Route multi-input flowsheets through sequential execution in `ProcessSystem.runOptimized` so reused mixers preserve mass balance and phase state.
- Add generic regression coverage for reused multiphase mixers.

The implementation and tests contain no application-specific GDR references.

## Validation and CI repair

Exact head: `f0a2b6883d976787d0d04b6077300a95def8de16`.

The first repair removed a duplicate `reconcileTotalMolesWithComponentInventory()` introduced while merging the branch with #3026. Fresh Java CI then exposed broad `SystemThermo` cache invalidation forcing empty and zero-flow systems through EOS initialization. That global invalidation was removed, leaving the change bounded to mixer phase bookkeeping, sequential process execution, and its focused regression test.

Exact-head CI confirms that the branch-attributable NaN cascade is gone: slow-test shards 1 and 4 pass, process serialization and distillation regressions pass, and the Windows fast suite reports only three merged-base failures. Spotless, pre-commit, PaperLab, CodeQL, and Javadocs pass.

**BLOCKED BY MERGED BASE REGRESSIONS:** the required Java workflow remains red only for failures introduced by merged #3026:

- `HydrocarbonScrubberSaturationPressureStabilityTest`: expected 105.9 bara, actual 72.24446868896484 bara.
- `SystemElectrolyteCPATest.testHydrateTemperatureWithMethanolAndSalt`: combined methanol+salt hydrate temperature is higher than methanol alone.
- `EclipseFluidReadWriteTest.testFluidWater3`: expected -4639.239569750378, actual -4646.270755989724.
- `TPflashIncipientPhaseStabilityTest.subResidualTpdDoesNotOverrideExistingUmrPruSolution`: expected one phase, actual two.
- `TPFlashTest.testRun5`: expected one phase, actual two.

These failures are outside this PR's mixer/process scope and are not repaired here.

Documentation impact: none — the repair removes unsafe internal cache invalidation and does not change public APIs, units, configuration, or documented usage.


EvenSol:
I tested the proposed changes against current `master` and opened a narrower alternative in #3047.

Key findings:

- The regression added in this PR passes on current `master` with both production changes removed.
- #3026 already reconciles stale `SystemThermo` total-mole inventory at Java deserialization, which is the boundary implicated by the restored-process failure.
- `ProcessGraphBuilder` creates producer dependencies for every mixer input, and `runParallel(UUID)` waits at topological level barriers before executing the mixer.
- Routing every multi-input flowsheet through `runSequential(UUID)` would remove independent-branch concurrency and the optimized dirty-unit skip path for mixers, manifolds, multi-stream heat exchangers, separators, tanks, and similar equipment.

#3047 is therefore test-only. It adds:

- a process-level serialize → restore → change MEG flow → `runOptimized()` regression;
- gas-phase, mass-balance, and component/total-inventory assertions; and
- an explicit assertion that multi-input equipment retains level-based parallel dispatch.

The focused and related suites pass (20 tests), as do Spotless, both pre-commit stages, and the documentation-search check.

My recommendation is to supersede this PR with #3047. If the original application still fails on current `master`, the next step should be a minimal serialized-process fixture and a fix at the deserialization or clone/state boundary—not a global scheduler fallback.

PR #3067, opened 2026-08-16
fix(pipeline): correct the stratified level solver and the Kelvin-Helmholtz criterion

EvenSol:
fix(pipeline): correct the stratified level solver and the Kelvin-Helmholtz criterion

Three defects in the horizontal flow-regime closure, found while chasing a step
change in hold-up across a regime transition.

1. The equilibrium-level solver never converged. estimateStratifiedLiquidLevel
   updated the level with a fixed step,

       h_L = h_L - 0.1 * D * Math.signum(resid);

   so from the 0.5*D initial guess it could only ever land on a 0.1*D grid, and
   its convergence test |h_L - h_prev| < 1e-6*D could never be met because the
   step was always exactly 0.1*D. It therefore ran all 20 iterations and
   returned an oscillating grid point. h_L/D jumped 0.4 -> 0.5 -> 0.6, so any
   criterion keyed on the level was a step function of gas velocity. Replaced
   with bisection on the same Taitel-Dukler residual, which is unchanged and
   was verified term by term against the published combined momentum equation.

2. The Kelvin-Helmholtz critical velocity was dimensionally wrong:

       U_G_crit = sqrt(dRho * g * h_G * A_G / (rho_G * S_i))

   The group under the root is m3/s2, not a velocity, because of the spurious
   extra h_G. Taitel and Dukler (1976) transition A is

       U_G >= C2 * sqrt(dRho * g * A_G / (rho_G * dA_L/dh_L)),  C2 = 1 - h_L/D

   with dA_L/dh_L = S_i. The coded form was roughly 0.3 to 0.55 times the
   correct critical velocity over the diameters of interest, so it declared the
   stratified layer unstable far too readily and over-called slug.

3. Closure was switched at a point rather than blended. A section on a
   transition is not wholly one regime, so hold-up and friction stepped as an
   operating point drifted across the boundary. PipeSection now carries
   fractional regime weights, FlowRegimeDetector.classify sets them from the
   Kelvin-Helmholtz margin and the equilibrium level ramped over a band, and
   TwoFluidPipe evaluates and weights each closure. The valley and peak terrain
   corrections likewise ramp in over the near-horizontal band instead of
   switching at 0.05 rad.

Reference-case behaviour is unchanged: pressure drop 10.65 / 35.43 / 79.58 /
138.70 bar and maximum hold-up 0.0221 are bit-identical to master, with and
without the equilibrium-level transition, because the corrected criterion
leaves that line stratified throughout.

The equilibrium-level transition stays opt-in. An earlier measurement appeared
to improve the low-rate hold-up deficit from -26% to -14%, but that gain came
from the incorrect criterion firing spuriously and does not survive the fix.
The hold-up deficit is therefore still open, as is the spurious pressure drop
across a zero-net-elevation undulation, which these changes showed to be a
closure question rather than a numerical one.

Pipeline suite: 818 tests, 0 failures, 22 skipped.

### DIFF
--- a/src/main/java/neqsim/process/equipment/pipeline/TwoFluidPipe.java
+++ b/src/main/java/neqsim/process/equipment/pipeline/TwoFluidPipe.java
@@ -1003,7 +1003,7 @@ public class TwoFluidPipe extends Pipeline {
           }
         }
 
-        logger.info("Three-phase flow detected: water cut = {:.1f}%, oil fraction = {:.1f}%", inletWaterCut * 100,
+        logger.info("Three-phase flow detected: water cut = {}%, oil fraction = {}%", inletWaterCut * 100,
             inletOilFraction * 100);
 
       } else if (hasOil || hasWater) {
@@ -1205,7 +1205,7 @@ public class TwoFluidPipe extends Pipeline {
     // Initialize accumulation tracker
     accumulationTracker.identifyAccumulationZones(sections);
 
-    logger.info("TwoFluidPipe initialized: {} sections, dx_min={:.2f}m{}", numberOfSections, dx,
+    logger.info("TwoFluidPipe initialized: {} sections, dx_min={}m{}", numberOfSections, dx,
         sectionLengths != null ? " (non-uniform mesh)" : "");
   }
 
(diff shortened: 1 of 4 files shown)


EvenSol:
Superseded by #3069, which contains these commits plus the separated-friction fix. Closing in favour of that one.

PR #3068, opened 2026-08-16
fix(logging): use Log4j2 placeholders instead of Python format specifiers

EvenSol:
fix(logging): use Log4j2 placeholders instead of Python format specifiers

40 logger calls across 11 files used Python-style format specifiers such as
{:.1f}, {:.2f} and {:.0f}. Log4j2 substitutes {} only, so these were emitted
literally and, wherever the specifier count differed from the argument count,
the remaining arguments bound to the wrong placeholders or were dropped.

The clearest example, in TwoFluidPipe:

    logger.warn("Steady-state solver reached wall-clock limit ({:.1f}s) after {} iterations",
        ssMaxWallClockTime, iter);

One real placeholder, two arguments: the message printed the configured timeout
where the iteration count belonged and discarded the iteration count. That
message is emitted exactly when a solve is truncated, so it misreported the one
case it exists to diagnose.

Replacing the specifiers with {} loses the intended decimal precision in the log
text, which is a fair trade for messages that substitute at all and bind their
arguments in the right order.

Found while investigating a truncated steady-state solve. No behaviour outside
logging is affected.

### DIFF
--- a/src/main/java/neqsim/process/equipment/compressor/CompressorWashing.java
+++ b/src/main/java/neqsim/process/equipment/compressor/CompressorWashing.java
@@ -360,7 +360,7 @@ public class CompressorWashing implements Serializable {
     hoursSinceLastWash = 0.0;
     lastWashMethod = method;
 
-    logger.info("Compressor wash performed: {} - Recovery: {:.1f}%", method.getDisplayName(), recovery * 100);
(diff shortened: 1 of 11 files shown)


EvenSol:
Superseded by #3069, which contains this commit. Closing in favour of that one.

PR #3088, opened 2026-08-17
fix: re-evaluate calculators during recycle convergence

### LINKED ISSUE #3083: Calculator not re-evaluating on recycles?
**Describe the bug**

`Calculator` units are evaluated once and never re-evaluated while a `Recycle`
loop iterates. The stale value keeps being injected on every iteration, so a
recycle loop that depends on a `Calculator` output grows linearly instead of
converging. No exception is raised and the recycle reports `solved() == false`,
but the process still returns a plausible-looking wrong answer.

This is **not** version-specific — it is latent on 3.1.5 as well. It only shows
when the `Calculator`'s first-pass value happens not to be self-consistent.

**To Reproduce**

A TEG dehydration loop where a `Calculator` computes glycol make-up as
`max(0, target − stripper outlet)` and feeds a make-up stream mixed back into
the lean solvent line:

```python
# Python example
# ... build TEG absorber -> flash -> regenerator -> stripper ...

makeup_teg = neqsim.process.equipment.stream.Stream("makeup TEG", pure_teg)
makeup_teg.setFlowRate(1e-6, "kg/hr")

makeup_calculator = neqsim.process.equipment.util.Calculator("Makeup calculator")
makeup_calculator.addInputVariable(stripper.getLiquidOutStream())
makeup_calculator.setOutputVariable(makeup_teg)


def calculate_makeup_teg(inputs, output):
    mass_flow_makeup = TARGET_FLOW - inputs[0].getFlowRate("kg/hr")   # TARGET_FLOW = 3800
    output.setFlowRate(max(0.0, mass_flow_makeup), "kg/hr")


makeup_calculator.setCalculationMethod(calculate_makeup_teg)
operations.add(makeup_calculator)

makeup_mixer = neqsim.process.equipment.mixer.StaticMixer("makeup mixer")
makeup_mixer.addStream(stripper.getLiquidOutStream())
makeup_mixer.addStream(makeup_teg)
# ... pump -> cooler -> lean TEG -> Recycle -> back to absorber ...

recycle_lean_teg = neqsim.process.equipment.util.Recycle("lean TEG recycle")
recycle_lean_teg.addStream(lean_TEG_to_absorber)
recycle_lean_teg.setOutletStream(teg_feed)
operations.add(recycle_lean_teg)

operations.run()
print(recycle_lean_teg.solved(), recycle_lean_teg.getIterations())
print(makeup_teg.getFlowRate("kg/hr"))          # identical at every iteration count
print(teg_feed.getFlowRate("kg/hr"))            # grows without bound
```

A complete runnable case is available on request (NeqSimAPI `SNA/TEG` model).

**Expected behavior**

The `Calculator` re-evaluates on each recycle iteration. As the loop flow
approaches the target, the computed make-up decays toward its converged value
and the recycle converges. This is what happens when the first-pass value is
already correct:

```
stripper liquid OUT   3799.7548  (TEG 3776.7971)
makeup TEG               0.2452  (TEG    0.2452)
makeup mixer OUT      3800.0000  (TEG 3777.0423)   <- closes exactly
```

**Actual behavior**

The `Calculator` output is **bit-identical** regardless of how many iterations
the recycle ran (14.217255 in both runs below), and the loop grows by exactly
one make-up injection per iteration:

| iterations | leanTEGflow (kg/hr) | growth ÷ makeup |
|---|---|---|
| 49 | 4507.52 | **49.76** |
| 100 | 5205.12 | **98.83** |

`flow = 3800 + n × 14.2173` exactly. The recycle terminates with
`solved=false`, `errorFlow` 0.27–0.31, and the process returns
`leanTEGflow = 5205 kg/hr` against a 3800 kg/hr setpoint — with no error raised.

**Environment:**
 - OS: Windows 11
 - Java version: JDK 25 (Eclipse Adoptium 25.0.4.7)
 - NeqSim version: 3.17.0 (exposed), 3.1.5 (latent — see below)
 - Python neqsim version (if applicable): jneqsim 3.17.0 / 3.1.5

**Additional context**

**Suspected root cause.** `ProcessSystem`'s iterative loop only re-runs
non-recycle units when `needsRecalculation()` is true:

```java
for (ProcessEquipmentInterface unit : plan.iterativeSection) {
    if (!(unit instanceof Recycle)) {
        if (iter == 1 || needsRecalculation(unit)) {
            runUnitProfiled(unit, id);
        }
    }
    ...
}
```

The default `needRecalculation()` returns `true`, so dirty-tracking is not the
problem — **placement** is. The units iterated come from `plan.iterativeSection`,
which is built from graph levels, and NeqSim's own source documents why a
`Calculator` cannot be placed there:

```java
/**
 * Calculator units read input streams and write to an output stream property using signal
 * connections rather than physical stream connections. They create implicit feedback loops
 * that the graph-based partitioner does not detect, so parallel execution cannot place
 * them correctly.
 */
public boolean hasCalculators() { ... }
```

A `Calculator` declares its dependency via `addInputVariable()` /
`setOutputVariable()` — a signal, not a stream — so the partitioner does not see
that it belongs inside the recycle loop, and it ends up in the feed-forward
section where it runs once.

**Why 3.17.0 exposes it.** The make-up is evaluated on the first pass:

| | first-pass stripper outlet | makeup | outcome |
|---|---|---|---|
| 3.1.5 | 3799.7548 | 0.2452 | self-consistent → loop stable at 3800 |
| 3.17.0 | 3785.78 | 14.2173 | not self-consistent → runaway |

On 3.1.5 the one-shot value happens to be right, so the loop is stable by luck.
3.17.0 returns ~14 kg/hr less TEG from the regeneration section on the first
pass, which is enough to trigger the runaway.

**Workarounds tried and rejected.**
`Recycle.setDownstreamProperty("flow rate")` masks it by overwriting the loop
flow each iteration:

```java
mixedStream.setFlowRate(outletStream.getFlowRate("kg/hr"), "kg/hr");
```

That cancels the accumulation, but the make-up stream then goes to zero flow and
the make-up mixer throws `PhaseSrkCPAs:init - Input totalNumberOfMoles must be
larger than or equal to zero` on every solve, and the reported make-up becomes
structurally zero rather than a physical quantity. Neither `setPriority(200)`
alone, `setMaxIterations(100)`, nor `AccelerationMethod.WEGSTEIN` changes the
outcome (all give a bit-identical 4507.52).

**Suggested diagnostic.** A warning when a recycle terminates with
`solved() == false` would have surfaced this immediately — currently the process
returns a wrong answer silently.

Sviatose:
## Summary
- route recycle flowsheets containing `Calculator` units through sequential execution so the calculator runs on every recycle pass
- retain hybrid execution for recycle flowsheets without calculators
- add a focused regression test that verifies a calculator callback is invoked more than once and reaches the recycle fixed point without `setDownstreamProperty("flow rate")`

Fixes #3083

## Verification
- `./mvnw test -Dtest=ProcessSystemTest#calculatorReevaluatesDuringRecycleConvergence`
- `./mvnw spotless:check`

### DIFF
--- a/src/main/java/neqsim/process/processmodel/ProcessSystem.java
+++ b/src/main/java/neqsim/process/processmodel/ProcessSystem.java
@@ -1498,11 +1498,10 @@ public class ProcessSystem extends SimulationBaseClass {
       // performs the active-state reset when runOptimized() is called directly.
       // When run() delegates here, its outer run scope already performed the
       // active-state reset, so repeating preparation in this dispatcher is wasteful.
-      if (hasAdjusters()) {
-        // Adjusters create implicit feedback loops via signal connections and
-        // iterate on a target variable. The graph partitioner cannot represent
-        // that iterative coupling, so adjuster-containing systems must run
-        // sequentially to ensure correct evaluation order.
+      if (hasAdjusters() || (hasRecycles() && hasCalculators())) {
+        // Adjusters and calculators create implicit feedback loops via signal
+        // connections. A calculator coupled to a recycle must be evaluated on
+        // every recycle pass, which requires insertion-order sequential execution.
         runSequential(id);
       } else if (hasRecycles()) {
         // Process has Recycle units. runHybrid() parallelises feed-forward levels
--- a/src/test/java/neqsim/process/processmodel/ProcessSystemTest.java
+++ b/src/test/java/neqsim/process/processmodel/ProcessSystemTest.java
@@ -47,6 +47,49 @@ public class ProcessSystemTest extends neqsim.NeqSimTest {
   ProcessSystem p;
   String _name = "TestProcess";
 
+  @Test
+  public void calculatorReevaluatesDuringRecycleConvergence() {
+    neqsim.thermo.system.SystemInterface methane = new neqsim.thermo.system.SystemSrkEos(298.15, 10.0);
+    methane.addComponent("methane", 1.0);
+
+    Stream recycleFeed = new Stream("recycle feed", methane);
+    recycleFeed.setFlowRate(1.0e-6, "kg/hr");
+    Stream makeup = new Stream("makeup", methane.clone());
+    makeup.setFlowRate(0.0, "kg/hr");
+
+    final int[] calculationCount = new int[] { 0 };
+    Calculator calculator = new Calculator("makeup calculator");
+    calculator.addInputVariable(recycleFeed);
+    calculator.setOutputVariable(makeup);
+    calculator.setCalculationMethod((inputs, output) -> {
+      calculationCount[0]++;
+      double requiredMakeup = 100.0 - ((Stream) inputs.get(0)).getFlowRate("kg/hr");
+      ((Stream) output).setFlowRate(Math.max(0.0, requiredMakeup), "kg/hr");
+    });
+
+    StaticMixer mixer = new StaticMixer("makeup mixer");
+    mixer.addStream(recycleFeed);
+    mixer.addStream(makeup);
+
+    Recycle recycle = new Recycle("makeup recycle");
+    recycle.addStream(mixer.getOutletStream());
+    recycle.setOutletStream(recycleFeed);
+    recycle.setTolerance(1.0e-4);
+
+    ProcessSystem process = new ProcessSystem();
+    process.add(recycleFeed);
+    process.add(makeup);
+    process.add(calculator);
+    process.add(mixer);
+    process.add(recycle);
+    process.run();
+
+    assertTrue(recycle.solved());
+    assertTrue(calculationCount[0] > 1);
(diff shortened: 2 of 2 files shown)


EvenSol:
Closing this PR as superseded by #3090.

#3090 fixes the missing `Calculator` stream dependencies in `ProcessGraphBuilder`, so calculators coupled to material recycles remain inside the hybrid iterative section. That addresses the root cause while preserving optimized execution.

This PR's flowsheet-wide sequential fallback for any process containing both a recycle and a calculator is broader than necessary and would disable hybrid scheduling even when the calculator is unrelated to the recycle.

Both focused regressions passed in GitHub Actions. The remaining shared CI failure is the unrelated `SevereSluggingExperimentalBenchmarkTest` failure in slow-test shard 1; it occurs with identical numerical results on both PRs.

Please continue review in #3090.

PR #3092, opened 2026-08-17
fix: make mixer inlet order invariant

### LINKED ISSUE #3085: mixer stream order
**Describe the bug**

`Mixer` / `StaticMixer` seed the mixed stream from `streams.get(0)` and force it
to two phases, so the result depends on the **order inlet streams were added**.
Adding a small solvent stream before the bulk gas stream produces a
non-deterministic phase split — the same flowsheet, run repeatedly in fresh
JVMs, sometimes loses the aqueous phase entirely.

Nothing is logged to indicate that stream order matters.

**To Reproduce**

Add a small MEG/water stream and a bulk gas stream to a `StaticMixer` in each
order, then read the mixed stream's phases. With the small stream first the
downstream aqueous phase intermittently disappears:

```python
# Python example
from jneqsim import neqsim

# ... build a water-saturated gas stream (bulk, ~10^5 kg/hr) and a lean MEG
#     stream (~10^-2 kg/hr) on SystemSrkCPAstatoil with setMixingRule(10) ...

mixer = neqsim.process.equipment.mixer.StaticMixer("gas/MEG mixer")
mixer.addStream(lean_MEG_stream)      # small stream FIRST -> non-deterministic
mixer.addStream(gas_stream)
operations.add(mixer)

# ... heater -> stream -> second heater -> three phase separator ...
operations.run()

fluid = operations.getUnit("inlet stream").getFluid()
print(fluid.getNumberOfPhases())
fluid.getPhase("aqueous")             # intermittently: RuntimeException
```

Swapping to bulk-gas-first makes it deterministic:

```python
mixer.addStream(gas_stream)           # bulk stream FIRST -> deterministic
mixer.addStream(lean_MEG_stream)
```

A complete runnable case is available on request (NeqSimAPI `GDR/MEGRate` model).

**Expected behavior**

The mixed stream is the sum of its inlets. The order in which streams were
attached should not change the result, and certainly should not make it
non-deterministic.

**Actual behavior**

Six fresh-JVM runs of the identical flowsheet, jneqsim 3.17.0:

| inlet order | correct 3-phase result |
|---|---|
| MEG first, gas second | **1 of 6** |
| gas first, MEG second | 6 of 6 |

On jneqsim 3.7.0 the MEG-first ordering gives 6 of 6, so the sensitivity is
newly exposed rather than newly introduced.

When it fails the downstream symptom is:

```
java.lang.RuntimeException: Phase with phase type AQUEOUS not found.
    at neqsim.thermo.system.SystemThermo.getPhase(SystemThermo.java:3084)
```

**Environment:**
 - OS: Windows 11
 - Java version: JDK 25 (Eclipse Adoptium 25.0.4.7)
 - NeqSim version: 3.17.0 (non-deterministic), 3.7.0 (deterministic)
 - Python neqsim version (if applicable): jneqsim 3.17.0 / 3.7.0

**Additional context**

**Suspect code.** `StaticMixer.run()` seeds from inlet 0 and forces two phases:

```java
// StaticMixer.java ~L103
mixedStream.setThermoSystem((streams.get(0).getThermoSystem().clone()));
mixedStream.getThermoSystem().setNumberOfPhases(2);
mixedStream.getThermoSystem().reInitPhaseType();
```

`Mixer` has the same asymmetry:

```java
// Mixer.java
L157: mixedStream = streams.get(0).clone(this.getName() + " mixed stream");
L496: SystemInterface thermoSystem2 = streams.get(0).getThermoSystem().clone();
L507: boolean inletMultiPhaseCheck = streams.get(0).getThermoSystem().doMultiPhaseCheck();
L508: SystemInterface thermoSystem2 = streams.get(0).getThermoSystem().clone();
```

L507 is notable: the mixed stream inherits the multiphase-check flag from
**inlet 0 only**, so whether multiphase detection runs depends on which stream
was added first.

`SimpleTEGAbsorber.run()` repeats the same pattern
(`mixedStream.setThermoSystem(streams.get(0)...clone())` then
`setNumberOfPhases(2)`), so any fix should be considered for that class too.

**Related second-solve issue (same flowsheet).** After fixing the inlet order,
the same model still failed unless every unit was re-solved once with the
flowsheet fully wired, immediately before each solve:

```python
for unit in operations.getUnitOperations():
    unit.run()
future = operations.runAsTask()
```

Without it the second heater fails with:

```
java.lang.IllegalArgumentException: Unexpected number of rows in B based on shape of A. Found=10 Expected=2
    -> Failed to run unit operation "GDR heater at SLP"
```

(10 components against a matrix sized for 2 phases — a phase/component structure
inconsistency.) Tested and rejected as alternatives: a single extra flowsheet
solve under the existing timeout, and running the pre-pass once after
construction instead of before every solve. Both still fail. The inconsistency is
regenerated by **every** `runAsTask()`, so it is not a one-time construction
artifact. This may be the same underlying defect or a separate one.

**Suggested diagnostic.** A warning when `streams.get(0)` is not the dominant
flow would make this class of problem self-evident — the flowsheet currently
gives no indication that inlet order is significant.

Sviatose:
## Summary
- select the largest active inlet as the `Mixer` thermodynamic template instead of inheriting setup from registration index 0
- add every non-template stream exactly once during mixing, retaining mass conservation for the selected template
- apply `Mixer.setMultiPhaseCheck(...)` as the explicit outlet configuration rather than inheriting it from an arbitrary inlet
- add a `StaticMixer` permutation regression with a low-flow MEG stream and bulk wet gas registered in both orders

Fixes #3085

## Verification
- `./mvnw test -Dtest=StaticMixerTest#testInletOrderDoesNotChangeDominantTemplateOrMultiphaseConfiguration`
- `./mvnw test -Dtest=StaticMixerTest`
- `./mvnw spotless:check`

### DIFF
--- a/src/main/java/neqsim/process/equipment/mixer/Mixer.java
+++ b/src/main/java/neqsim/process/equipment/mixer/Mixer.java
@@ -180,6 +180,15 @@ public class Mixer extends ProcessEquipmentBaseClass implements MixerInterface, 
    * mixStream.
    */
   public void mixStream() {
+    mixStream(streams.get(0));
+  }
+
+  /**
+   * Mixes every inlet except the stream already represented by the mixed-stream template.
+   *
+   * @param templateStream inlet stream cloned as the mixed-stream template
+   */
+  private void mixStream(StreamInterface templateStream) {
     int index = 0;
     // Determine outlet pressure from the lowest pressure across ACTIVE inlet streams only.
     // Streams with negligible flow (bypassed / deactivated) must not influence the mixer
@@ -222,9 +231,11 @@ public class Mixer extends ProcessEquipmentBaseClass implements MixerInterface, 
       }
     }
 
-    // Process ALL streams starting from k=1 (k=0 is already cloned into mixedStream)
-    // but ensure first stream's components are also explicitly added if needed
-    for (int k = 1; k < streams.size(); k++) {
+    // The template stream is already present in mixedStream; add every other inlet exactly once.
+    for (int k = 0; k < streams.size(); k++) {
+      if (streams.get(k) == templateStream) {
+        continue;
+      }
       // Skip streams with negligible flow to avoid mixing in zero/negative moles
       if (streams.get(k).getFlowRate("kg/hr") <= getMinimumFlow()) {
         continue;
@@ -293,6 +304,33 @@ public class Mixer extends ProcessEquipmentBaseClass implements MixerInterface, 
     if (hasAddedNewComponent) {
       mixedStream.getThermoSystem().setMixingRule(mixedStream.getThermoSystem().getMixingRule());
     }
+  }
+
+  /**
+   * Selects a stable thermodynamic template from the active inlet streams.
+   *
+   * <p>
+   * The largest active mass-flow stream is used so a small injection stream cannot control the outlet's thermodynamic
+   * configuration merely because it was registered first. Equal-flow ties are resolved by stream name, making the
+   * selection independent of registration order.
+   * </p>
+   *
+   * @return the selected thermodynamic template stream
+   */
+  private StreamInterface selectTemplateStream() {
+    StreamInterface templateStream = streams.get(0);
+    double maximumFlow = Double.NEGATIVE_INFINITY;
+    for (StreamInterface stream : streams) {
+      double flow = stream.getFlowRate("kg/hr");
+      if (flow <= getMinimumFlow()) {
+        continue;
+      }
+      if (flow > maximumFlow || (flow == maximumFlow && stream.getName().compareTo(templateStream.getName()) < 0)) {
+        templateStream = stream;
+        maximumFlow = flow;
+      }
+    }
+    return templateStream;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


EvenSol (changes requested):
The dominant-flow selection fixes the reported two-inlet case, but this is not yet generally inlet-order invariant and it changes multiphase performance semantics:

1. `thermoSystem2.setMultiPhaseCheck(doMultiPhaseCheck)` forces multiphase checking on for every default `Mixer`, even when all active inlets have it disabled. The current mixer flag is an allow/disable override; the order-independent replacement should OR the active inlet flags and still honor `Mixer.setMultiPhaseCheck(false)`.
2. Non-template streams are still accumulated in registration order, so 3+ inlet component creation and floating-point accumulation remain order dependent.
3. The all-inactive path still clones `streams.get(0)`.
4. Equal-flow ties use only stream name; duplicate names can still select different thermodynamic templates, and no regression verifies equal-flow systems with different mixing-rule configurations.
5. The user-visible template behavior is not documented.

#3093 contains the complete variant and stronger evidence: all six permutations of gas/MEG/water for both `Mixer` and `StaticMixer`, a second solve after changing MEG flow, component/phase/inventory comparisons, and an equal-flow regression that fails on `master` with `NO` versus `CLASSIC` template rules. It retains one flash and optimized parallel execution. I recommend closing this PR in favor of #3093 or incorporating those changes here.

PR #3224, opened 2026-08-24
Resolve component names across naming conventions; fix component lookup SQL and COMP_EXT superset

dupuypablo:
## Summary

Three related fixes to how a component name reaches the NeqSim component database, found while
cross-checking a laboratory DHA (ASTM D6729) composition report against `COMP.csv`. Of 245
unambiguously identified components in that report, only 74 could be added to a fluid — not because
the parameters were missing, but because the names did not match. Most were reachable under a
different spelling.

| | |
|---|---|
| **1. Name resolution** | New `ComponentNameResolver`; `addComponent` now accepts systematic, trivial, case-varied and inverted-CAS names |
| **2. Component lookup** | `hasComponent` / `hasTempComponent` now use a prepared statement |
| **3. Extended database** | `COMP_EXT.csv` is now genuinely a superset of `COMP.csv`, as documented |

No behaviour that worked before changes. `ComponentInterface.getComponentNameFromAlias(String)`
keeps its signature and delegates to the resolver, so all ~28 existing call sites benefit without
modification.

---

### The problem

`addComponent(...)` mapped names through a hardcoded 18-entry `LinkedHashMap` of reservoir
shorthand in `ComponentInterface.getComponentNameMap()`:

```
H2O→water  N2→nitrogen  C1→methane  C2→ethane  C3→propane  iC4→i-butane
nC4→n-butane  iC5→i-pentane  nC5→n-pentane  C6→n-hexane  nC7→n-heptane
nC8→n-octane  nC9→n-nonane  O2→oxygen  He→helium  H2→hydrogen  Ar→argon  H2S→H2S
```

(`O2` was inserted twice, so 17 distinct entries.) Lookup was `containsKey` — exact and
case-sensitive — and the database query behind it is `WHERE NAME='...'` on H2 with connection
string `jdbc:h2:mem:neqsimthermodatabase`, i.e. no `IGNORECASE=TRUE`. So the match was
case-sensitive end to end.

That map contains no chemical synonyms at all. The database stores 2,2,4-trimethylpentane as
`224-TM-C5`, so every one of these failed with *"not found in database"* even though the component
is present:

```java
fluid.addComponent("2,2,4-trimethylpentane", 1.0);  // systematic name
fluid.addComponent("isooctane", 1.0);               // trivial name
fluid.addComponent("ISOOCTANE", 1.0);               // wrong case
fluid.addComponent("Cyclohexane, 1,2,4-trimethyl-", 1.0);  // CAS index name from a GC report
```

The last form matters in practice: chromatography software mixes prefix-style names
(`2-Methylpentane`) with inverted CAS index names (`Cyclopentane, 1-ethyl-2-methyl`) in the same
table. The report that prompted this used the inverted form for 19 of its 402 peaks.

### The change

New `neqsim.thermo.component.ComponentNameResolver`. Resolution proceeds in four steps, first hit
wins:

1. case-insensitive match against a database name (257 entries);
2. match against the synonym table (148 entries);
3. the same two lookups after converting an inverted CAS index name to prefix form;
4. otherwise the input is returned unchanged.

| Convention | Example | Resolves to |
|------------|---------|-------------|
| Database name, any case | `N-HEPTANE` | `n-heptane` |
| Reservoir shorthand (as before) | `C1`, `nC4`, `iC5`, `H2O` | `methane`, `n-butane`, `i-pentane`, `water` |
| Systematic name | `2,2,4-trimethylpentane` | `224-TM-C5` |
| Trivial name | `isopentane`, `neopentane`, `cyclohexane` | `i-pentane`, `22-dim-C3`, `c-hexane` |
| Inverted CAS index name | `Cyclohexane, 1,2,4-trimethyl-` | `1.2.4-TMcyC6` |
| Either locant separator | `1,2,3-TM-Benzene` / `1.2.3-TM-Benzene` | `1.2.3-TM-Benzene` |

### How the synonym table was built

It is **derived from `COMP.csv` itself**, not hand-written. A script parses each database name out
of its in-house shorthand grammar (`224-TM-C5` → 2,2,4-trimethylpentane) and emits the systematic
name as a synonym pointing back at the database name. 62 entries come from that expansion; the
remaining 86 are curated trivial and IUPAC names.

**Ambiguous expansions are skipped, not guessed.** Five shorthand forms expand to a structure that
more than one database entry shares:

```
1,3-dimethylcyclopentane → 1.3-dDM-cyC5, cis-13-DM-cy-C5, trans-13-DM-cy-C5
1,2-dimethylcyclopentane → 1.2-DM-cyC5, trans-12-DM-cy-C5
1,2-dimethylcyclohexane  → cis-12-DM-cy-C6, trans-12-DM-cy-C6
1,4-dimethylcyclohexane  → cis-14-DM-cy-C6, trans-14-DM-cy-C6
2-butene                 → cis-butene, trans-butene
```

These parent names are deliberately absent from the table, so an under-specified name is passed
through rather than silently resolved to one of the partners:

```java
ComponentNameResolver.resolve("cis-1,3-dimethylcyclohexane"); // "cis-13-DM-cy-C6"
ComponentNameResolver.resolve("1,3-dimethylcyclopentane");    // unchanged
ComponentNameResolver.isKnownName("1,3-dimethylcyclopentane"); // false
```

Stereochemistry given by the caller is always honoured. Stereochemistry the caller did not give is
never invented.

The table build also fails loudly on a duplicate key (`put` throws `IllegalStateException`), so an
accidental collision cannot ship silently.

### Compatibility

- `getComponentNameFromAlias(String)` keeps its signature; every previously accepted name resolves
  to the same component. Covered by `legacyReservoirShorthandStillResolves`.
- `getComponentNameMap()` is deprecated in favour of `ComponentNameResolver.getSynonyms()`. It
  still returns the original 18 entries.
- Unknown names are returned unchanged, so components that exist only in the extended database are
  unaffected.
- The tables target the standard database. This is stated in the class JavaDoc.

---

## 2. SQL injection in `hasComponent` / `hasTempComponent`

`NeqSimDataBase.hasComponent` assembled its query by string concatenation:

```java
"select count(*) from comp WHERE NAME='" + name + "'"
```

This is the classic injection pattern, but it is also a **live functional bug**: component names
legitimately contain apostrophes. **2,061 names in `COMP_EXT.csv` contain one** — for example
`4'-hydroxyacetophenone`, `n,n'-diphenyl-p-phenylenediamine`, `4,4'-diisopropylbiphenyl`. Every one
of those produces invalid SQL, so those components can never be looked up. `COMP.csv` contains no
such name, which is why this has never surfaced.

Both methods now delegate to a private `countRowsByName(table, name)` that binds the name with
`PreparedStatement.setString`, and return `false` for `null` instead of reaching the database. The
table name is a class-internal literal, not caller input.

`NeqSimDataBaseComponentLookupTest.hasComponentToleratesQuotesInTheName` asserts that
`"' OR '1'='1"` and `"x'; DROP TABLE comp; --"` return `false` without throwing, and that the
`comp` table is still intact afterwards.

---

## 3. `COMP_EXT.csv` was not a superset of `COMP.csv`

`docs/thermo/component_list.md` stated:

> *The extended database contains all components from the standard database (with identical
> parameters) plus over 50,000 additional components*

It did not. **Eleven standard components were missing:**

```
NO2-  H2O2  NaNO2  NH2SO3H  NH2SO3-  NaHSO4  N2O4  N2O4-  asphaltene  MEA+  MEACOO-
```

This matters because `useExtendedComponentDatabase(true)` **drops and recreates** the `COMP` table
rather than adding to it. Calling it silently removed these components from every fluid built
afterwards, and it is a global static, so the effect is process-wide. `asphaltene` and the MEA
species are not obscure.

The eleven rows are copied verbatim from `COMP.csv` (identical parameters, as documented) with `ID`
and `COMPINDEX` continuing the `COMP_EXT` sequence. The relationship is now enforced by
`extendedDatabaseContainsEveryStandardComponent`, which reads both CSVs from the classpath.

> Note for reviewers: `COMP_EXT.csv` is 82.7 MB, so GitHub prints a large-file warning on push. The
> change is an 11-line append and deltas well, but if you would rather not touch that file, the
> alternative is to correct the documentation instead and leave the eleven components missing. I
> chose to make the documented behaviour true because the failure mode — components vanishing from
> a global static switch — is silent.

---

## Documentation

`docs/thermo/component_list.md`:

- **New "Component Name Resolution" section** — the accepted conventions, the stereochemistry rule,
  and the introspection API. Every code example is asserted in
  `ComponentNameResolverTest.documentedExamplesBehaveAsDescribed`.
- **Corrected component counts** — 257 standard and 76,704 extended; the page claimed "~250" and
  "50,000+".
- **Superset claim** — reworded to explain *why* it matters (the table is replaced, not merged) and
  to reference the test that enforces it.
- **Provenance warning added.** The origin of the ~76,000 extra components in `COMP_EXT.csv` is not
  documented anywhere in the repository, and the entries have not been validated against a primary
  source such as DIPPR 801 or NIST. The page now says so and advises checking parameters taken from
  it. I could not establish provenance — **if a reviewer knows where this file came from, it would
  be worth recording.**

`CHANGELOG_AGENT_NOTES.md`: entry describing all three changes and the deprecation.

---

## Tests

New, 18 assertions across two classes:

**`ComponentNameResolverTest`** (14) — reads `data/COMP.csv` from the classpath so the tests fail if
the resolver tables drift away from the database rather than from a copy of it:

- every database name resolves to itself, and in upper and lower case;
- every synonym target exists in the database;
- the legacy reservoir shorthand still resolves, directly and through `getComponentNameFromAlias`;
- systematic, trivial and inverted-CAS names resolve;
- stereochemically ambiguous names are *not* resolved; explicit stereochemistry is;
- unknown names, `null`, empty string and surrounding whitespace behave;
- the documented examples behave as documented;
- an end-to-end check that `SystemSrkEos.addComponent("2,2,4-trimethylpentane", …)` builds the
  component named `224-TM-C5`.

**`NeqSimDataBaseComponentLookupTest`** (4) — known/unknown lookup, apostrophes and injection
attempts, `null` handling, and the `COMP_EXT` superset check.

```
Tests run: 18, Failures: 0, Errors: 0, Skipped: 0
```

`spotless:apply` has been run. All code is Java 8 compatible — explicit types throughout, no `var`,
no `List.of`/`Map.of`, no text blocks; the one anonymous `Executable` in the database test is
written as an anonymous class rather than a lambda for consistency with the surrounding style.

---

## Not included

The original motivation was to add missing components to `COMP.csv`. That is **deliberately not in
this PR.** The cross-check found 171 report components absent from the standard database, of which
123 match structurally against `COMP_EXT.csv` and most of the remainder are present there under
trivial names. Promoting them needs decisions this PR should not make silently:

- 18 cases where the structural match is right but the cis/trans partner must be chosen by hand;
- a naming convention decision (keep the `COMP_EXT` name, or rename to the `2-M-C6` shorthand);
- and `COMP_EXT`'s provenance is unresolved, so its parameters should be spot-checked against a
  primary source first.

Given that `COMP.csv` already carries placeholder data — CAS `74-82-8` (methane) on 71 rows,
`NORMBOIL = 58.05 °C` on 33 hydrocarbons including `nC39` and ethylbenzene, and `2-M-C8` stored
with the molar mass of a C7 — bulk-importing 160 unverified rows would make that worse. Better as a
separate, reviewed change. Happy to follow up if the approach here looks right.


EvenSol (approved):
good fix

PR #3242, opened 2026-08-25
Make the straight-chain n- prefix optional in ComponentNameResolver

dupuypablo:
## Summary

Follow-up to #3224. The synonym table generated there left an inconsistent cliff at C10, so
`addComponent("n-decane")` failed on a component the database does have.

```java
ComponentNameResolver.resolve("n-nonane");  // "n-nonane"   OK
ComponentNameResolver.resolve("n-decane");  // "n-Decane"   not found
```

## Cause

The database spells straight-chain alkanes two different ways:

| Carbon number | Stored as |
|---|---|
| C4 – C9 | `n-butane`, `n-pentane`, … `n-nonane` |
| C10 and up | `nC10`, `nC11`, … `nC24`, `nC30`, `nC39` |

The synonym table is generated by expanding the in-house shorthand, so `nC10` produced the synonym
`decane`. Nothing produced `n-decane`. C4–C9 happened to work only because those *are* the literal
database names.

So the obvious name worked up to C9 and then silently stopped — the worst kind of boundary, because
nothing about `n-decane` looks unusual to a caller.

## Change

`resolve()` retries with a leading `n-` removed, **but only after a direct lookup has missed**:

```java
String hit = lookup(key);
if (hit != null) {
  return hit;                      // 'n-butane' matches here, untouched
}
if (key.startsWith("n-")) {
  hit = lookup(key.substring(2));  // 'n-decane' -> 'decane' -> 'nC10'
  ...
}
```

Ordering matters: because the direct lookup runs first, every database name that legitimately begins
with `n-` is matched as-is and never goes through the strip. `isKnownName` follows the same path.

The `n-` prefix only marks a chain as unbranched and carries no information the database name needs,
so removing it cannot change which molecule is meant.

## Effect

Measured end to end against the 245 uniquely identified components of an ASTM D6729 DHA report —
each label resolved and then looked up with `hasComponent`:

| | Before | After |
|---|---:|---:|
| `addComponent` works on the standard database | 55 | **63** |

The eight recovered are the six C10+ n-alkanes (`n-Decane`, `n-Undecane`, `n-Dodecane`,
`n-Tridecane`, `n-Tetradecane`, `n-Pentadecane`) plus `n-Propylbenzene` → `propylbenzene` and
`n-Hexylbenzene` → `nC6-Benzene`, where the same prefix was blocking an aryl substituent.

## Tests

New `straightChainPrefixIsOptional` in `ComponentNameResolverTest`:

- `n-decane`, `n-undecane`, `n-dodecane`, `n-pentadecane`, `n-icosane` resolve to `nC10`…`nC20`;
- `decane` without the prefix still resolves;
- `n-pentylbenzene` → `nC5-Benzene`;
- `n-butane` and `n-nonane` are returned **unchanged**, guarding the ordering above;
- `isKnownName("n-decane")` is true.

```
Tests run: 20, Failures: 0, Errors: 0, Skipped: 0
```

The existing `databaseNamesResolveToThemselves` test (every one of the 257 database names must
resolve to itself, in three letter cases) already covers the regression risk of adding a fallback.

`spotless:apply` run. Java 8 compatible.


EvenSol (changes requested):
The C10 naming cliff is real, but I do not think a resolver-wide `n-` stripping fallback is safe.

The new branch accepts `n-` in front of **every** canonical name or synonym, not only a straight-chain name. For example, on this head:

- `resolve("n-water")` returns `"water"`;
- `resolve("n-acetone")` returns `"acetone"`;
- more seriously, `resolve("n-isobutane")` returns `"i-butane"`;
- `isKnownName(...)` reports all of these as known.

That breaks the documented invariant that unknown names are returned unchanged and can silently select a chemically different component. Doing the direct lookup first protects existing canonical names such as `n-butane`, but it does not constrain the stripped lookup.

Please keep this in the alias/synonym layer instead. A bounded option is to add `n-` variants only for synonym entries proven to describe a normal chain—for example, entries whose canonical target matches the NeqSim straight-chain form `nC\\d+(?:$|-.*)`. That covers `decane -> nC10`, `pentylbenzene -> nC5-Benzene`, and the other generated normal-chain rows without making `n-` a universal ignorable prefix. Explicit paired aliases via a helper would also be fine and even more conservative. The existing curated `n-propanol`, `n-propylbenzene`, and `n-butylcyclohexane` entries can remain explicit.

Please retain the positive tests and add negative regressions such as:

```java
assertEquals("n-water", ComponentNameResolver.resolve("n-water"));
assertEquals("n-isobutane", ComponentNameResolver.resolve("n-isobutane"));
assertFalse(ComponentNameResolver.isKnownName("n-water"));
assertFalse(ComponentNameResolver.isKnownName("n-isobutane"));
```

Because this changes accepted user input, please also update the nearby component-name-resolution documentation (or record an evidence-based documentation-impact assessment).

EvenSol:
## Repair update

Exact head: `720daaefcd7b4221aaeb0c3aa43410d4afd0b73e`

The requested-changes finding is addressed without a resolver-wide prefix strip:

- `n-` variants are generated only for existing aliases whose canonical target is an explicit normal-chain `nC...` component;
- positive coverage retains `n-decane`, `n-undecane`, `n-dodecane`, `n-pentadecane`, `n-icosane`, and `n-pentylbenzene`;
- negative regressions keep `n-water`, `n-acetone`, `n-isobutane`, and `n-n-decane` unknown and unchanged;
- the component-name guide now documents the bounded alias rule;
- the existing branch-wide Spotless violation in `McpEvidenceInventory.java` was formatted so the pre-commit gate can pass.

Local validation on the published source:

- `python devtools/run_spotless.py apply` — passed and stable;
- `./mvnw -q -DskipITs -Dtest=ComponentNameResolverTest test` — passed;
- `./mvnw -q --file pomJava8.xml -DskipITs -Dtest=ComponentNameResolverTest test` — passed;
- `python devtools/run_spotless.py check` — passed;
- `python devtools/check_documentation_search.py` — passed (659 Markdown pages, 1 standalone HTML page);
- `pre-commit run --all-files --hook-stage pre-commit` — passed;
- `pre-commit run --all-files --hook-stage pre-push` — passed;
- `git diff --check` — passed.

Documentation impact: updated `docs/thermo/component_list.md` and exercised its resolver examples in `ComponentNameResolverTest`.

**VALIDATION PENDING CI:** local Javadoc could not run because this container has a Java 17 runtime without `javac`/`javadoc`; hosted Javadoc and cross-platform suites remain authoritative. No unrun check is claimed as passed.

EvenSol (approved):
The requested safety boundary is addressed on the exact head: `n-` aliases are generated only for canonical `nC...` normal-chain targets, negative regressions keep `n-water`, `n-acetone`, `n-isobutane`, and `n-n-decane` unknown, and the nearby component-name documentation records the behavior. Local validation: `ComponentNameResolverTest` 16/16 passed; Spotless check and documentation-search audit passed.

EvenSol:
## Conflict resolution update

Resolved and published on exact head `7f57919c3d2b516736a81e357aab82a70215a12d`.

- merged current `master` parent `2b330659716d157727fd88ebc5d7c8019d88aa86` into the existing PR branch;
- resolved the sole conflict in `McpEvidenceInventory.java` by retaining `master`'s already-promoted empty contract-candidate inventory and dropping the PR's obsolete formatting-only version;
- preserved the bounded normal-chain alias implementation, focused regression tests, and component-name documentation;
- verified the final PR diff contains only the intended three files;
- GitHub now reports the PR as mergeable with no unresolved review threads. The two newer `master` commits touch disjoint files.

Exact-head local validation:

- `./mvnw -q -DskipITs -Dtest=ComponentNameResolverTest test` — passed (16 tests, 0 failures/errors);
- `./mvnw -q --file pomJava8.xml -DskipITs -Dtest=ComponentNameResolverTest test` — passed;
- `python3 devtools/run_spotless.py apply` — passed and stable;
- `python3 devtools/run_spotless.py check` — passed;
- `python3 devtools/check_documentation_search.py` — passed (674 Markdown pages, 1 standalone HTML page);
- `pre-commit run --all-files --hook-stage pre-commit` — passed;
- `pre-commit run --all-files --hook-stage pre-push` — passed;
- final diff check and conflict-marker scan — passed.

Documentation impact: the existing update to `docs/thermo/component_list.md` remains accurate and is covered by the focused resolver regression test.

**VALIDATION PENDING CI:** six exact-head workflows are currently in progress; no unrun hosted check is claimed as passed.

PR #3372, opened 2026-08-31
Add separator internals elevation API and gas scrubber drainage head check

dupuypablo:
## What

Adds the internals elevation API on `SeparatorMechanicalDesign`, plus the derived
cyclone-top elevation and drainage-head check on `GasScrubberMechanicalDesign`.

This is the public geometry an entrainment / carry-over correlation needs to read
in order to know how far a droplet travels between internals. It is a
prerequisite for pluggable carry-over models; the model itself is not part of
this PR.

## Reference frame

All elevations are **metres from the vessel Bottom Tangent Line (BTL), positive
upward**. This is stated on every accessor, because an elevation without a datum
is the obvious place for a silent error.

For each internal we store the **gas-facing surface** most relevant to free-path
calculations:

| Internal | Stored surface |
|---|---|
| Inlet nozzle | centreline elevation |
| Inlet device | TOP face (the surface gas leaves through) |
| Mesh pad / vane pack | BOTTOM face (the surface gas enters through) |
| Demisting cyclones | deck elevation = BOTTOM of the tubes |

## API added

`SeparatorMechanicalDesign`

- `set/getInletNozzleElevation(double)` — nozzle centreline [m]
- `set/getInletDeviceTopElevation(double)` — device top face [m]
- `getInletDeviceTopElevationOrDefault()` — falls back to
  `inletNozzleElevation + inletNozzleID / 2` when not explicitly set, matching the
  standard inlet-vane assumption that the device fills the upper half of the bore
- `set/getMeshPadBottomElevation(double)` — pad bottom face [m]
- `getMeshPadTopElevation()` — bottom plus `getDemisterThickness()`, which is
  stored in **mm** and converted here
- `getFreePathHeightAboveInletDevice()` — mesh pad bottom minus inlet device top

`GasScrubberMechanicalDesign`

- `getCycloneTopElevation()` — `cycloneDeckElevationM + cycloneLengthM`
- `computeDrainageHead()` — returns a `DrainageHeadResult`

`DrainageHeadResult` (new, immutable, `Serializable`)

## Drainage head

The liquid film draining from the cyclone bank back into the bulk must balance

```
dP_total = dP_mesh + dP_cyclone_to_drain
```

where `dP_mesh = Eu · 0.5 · rho_g · v_mesh²` (Euler coefficient defaults to 0.5
for standard knitmesh when unset) and the cyclone share reaching the drain
chamber is set by `cycloneDpToDrainPct`. The equivalent clear-liquid column
accounts for gas buoyancy:

```
h_required = dP_total / ((rho_L − rho_G) · g)
```

Available head is the geometric distance from the cyclone deck down to LA(HH).
`getPercentOfAvailable()` is the governing metric — above 100 % the drainage
column floods the cyclone bank.

## Behaviour and compatibility

Purely additive — **no existing line is modified** (572 insertions, 0 deletions).
Unset elevations return `0.0`, and every derived getter guards on that, so no
existing model changes behaviour. Inverted geometry (mesh pad below the inlet
device) returns `0.0` rather than a negative free path.

Java 8 compatible. `DrainageHeadResult` is `Serializable` with a
`serialVersionUID`, consistent with the other mechanical-design result classes.

## Tests

`SeparatorInternalsElevationTest` — 9 tests asserting exact values rather than
non-null:

- accessor round-trip and zero defaults
- inlet-device-top fallback = 1.35 m for a 1.20 m nozzle at 0.30 m bore, and that
  the fallback requires *both* inputs
- mesh pad top = 2.40 m for a 2.25 m bottom and 150 mm pad, covering the mm→m
  conversion
- free path = 0.90 m, and the unset / inverted-geometry guards
- cyclone top = 3.30 m for a 2.50 m deck and 0.80 m tube, and that it needs both
- drainage head self-consistency: total dP equals the sum of its parts, available
  head follows deck−LA(HH), the percentage equals required/available, and the
  liquid is denser than the gas

## Verification status

Please rely on CI for the build. The author's workstation currently has no
runnable JDK — AppLocker blocks executables under the user profile, where the
only installed JDK and Maven live — so `spotless:apply`, `compile` and `test`
could not be run locally. Formatting was matched to the surrounding code by hand;
if `spotless:check` objects I will push the fixup.

## Origin

Extracted from the long-lived `scrubbers_kollsnes` branch and re-applied onto
current `master`. That branch was 4 months behind and is superseded.

### DIFF
--- a/src/main/java/neqsim/process/mechanicaldesign/separator/DrainageHeadResult.java
+++ b/src/main/java/neqsim/process/mechanicaldesign/separator/DrainageHeadResult.java
@@ -1,0 +1,141 @@
+package neqsim.process.mechanicaldesign.separator;
+
+import java.io.Serializable;
+
+/**
+ * Breakdown of the demisting-cyclone drainage head check for a gas scrubber.
+ *
+ * <p>
+ * Represents the total pressure drop the liquid film must overcome to drain
+ * from the cyclone bank back into the vessel bulk — mesh pad pressure drop
+ * plus the fraction of cyclone pressure drop acting at the drain chamber —
+ * expressed as an equivalent clear-liquid column height. The available head
+ * is the geometric elevation from the cyclone deck down to the High-High
+ * Liquid Level (LA(HH)). The ratio of required to available head (percent)
+ * is the governing conformity metric.
+ * </p>
+ *
+ * @author NeqSim Development Team
+ * @version 1.0
+ */
+public class DrainageHeadResult implements Serializable {
+  /** Serialization version UID. */
+  private static final long serialVersionUID = 1000L;
+
+  private final double meshDpPa;
+  private final double cycloneDpToDrainPa;
+  private final double totalDpPa;
+  private final double requiredHeadMm;
+  private final double availableHeadMm;
+  private final double percentOfAvailable;
+  private final double gasDensityKgM3;
+  private final double liquidDensityKgM3;
+
+  /**
+   * Constructs a DrainageHeadResult.
+   *
+   * @param meshDpPa             mesh pad pressure drop [Pa]
+   * @param cycloneDpToDrainPa   cyclone pressure drop reaching drain chamber [Pa]
+   * @param requiredHeadMm       equivalent clear-liquid head required [mm]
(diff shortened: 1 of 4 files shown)


EvenSol (approved):
Good update

PR #3443, opened 2026-09-04
Add ConstraintSource provenance axis and EmpiricalCarryOverConstraint

dupuypablo:
## What

Completes the **two-axis capacity-constraint model**. The severity axis
(`ConstraintSeverity` — *how* a violation is handled) is already on master. This
adds the provenance axis (`ConstraintSource` — *what authority backs the limit*),
plus the first constraint type that depends on it.

A vendor surge limit and a company K-factor may both be `HARD`, but they are not
renegotiable on the same terms. Recording the source lets a capacity report state
both *what limits production today* and *what would limit it if conformity were
enforced*.

### `CapacityConstraint`
- New `ConstraintSource` enum: `CONFORMITY_STANDARD`, `USER_RULE`,
  `PROCESS_EMPIRICAL`, `VENDOR_DATASHEET`, `AUTO_SIZE`, `DEFAULT`.
- `setSource(source)`, `setSource(source, reference)`,
  `setSourceReference(reference)`, `getSource()`, `getSourceReference()`.
- All additive; default is `DEFAULT`, so existing constraints are unaffected.

### `EmpiricalCarryOverConstraint` (new)
Packages the common field situation where the binding limit is **not** a
standard's K-factor but an observed downstream consequence — liquid carry-over
from the last scrubber accumulating in a downstream compressor suction drum. It
is a piecewise-linear correlation calibrated against measured points, and an
ordinary `CapacityConstraint`, so it participates in bottleneck detection, the
utilization snapshot, and the optimiser.

### `ProcessSystem`
`getUtilizationSnapshot()` now emits `source` and `sourceReference` per
constraint alongside the existing `dataSource`, so the new axis is visible to
agents and reports rather than being write-only.

### `Separator` — JavaDoc only, no behaviour change
`designLiquidLevelFraction` was documented **twice** as "vertical only". The code
derates the gas area by `(1 - fraction)` only for **horizontal** separators
(lines 1437, 1655, 1926, 2097, 2146); a vertical separator uses the full
cross-section. Both statements corrected, plus the getter, which documented
`@return the designGasLevelFraction` — the wrong field.

### Docs
New sections in `docs/process/equipment/separators.md`: constraint provenance,
empirical carry-over constraints, and choosing which layer binds.

## Correctness fixes made while migrating

These are worth flagging because they change behaviour relative to the draft this
was migrated from:

1. **`ConstraintType.HARD`, not `SOFT`.** `isHardLimitExceeded()` returns `false`
   unless `type == ConstraintType.HARD`. The draft constructed the constraint as
   `SOFT` while setting severity `HARD` and documenting it as "hard, binding", so
   a carry-over blow-past would never have raised the plant-wide
   `anyHardLimitExceeded` feasibility gate.
2. **Single-calibration-point guard.** With one point, the extrapolation branch
   indexed `yPoints[-1]`. It now degenerates to a constant.
3. **Unit contract stated explicitly.** The class does not interpret units; the
   JavaDoc states that the driver must match `xPoints`, and that `yPoints` and
   `maxAllowable` share the constraint unit.

## Tests

`EmpiricalCarryOverConstraintTest` — hand-computed expected values (no
"looks reasonable" assertions): clamping below range, exact reproduction at every
knot, linear interpolation, last-segment extrapolation, utilization, hard-limit
tripping, provenance defaults, defensive copying on input and output, rejection
of null/empty/mismatched/non-ascending arrays, single-point degeneration, and
registration on a `Separator`.

## For the reviewer

- **`ConstraintSource` vs the existing `dataSource`.** These overlap and I want a
  second opinion on the split. My reading: `dataSource` is a free-text label for
  *which part of the model* supplied the number (`"equipment"`,
  `"mechanicalDesign"`, `"installed_capacity_table"`) and is already consumed in
  ~40 places (JSON snapshot, `PlantConstraintRegistry.provenance(...)`,
  debottleneck CSV, the `"not_set"` checks). `ConstraintSource` classifies the
  *authority* behind the number. I kept them independent and documented the
  distinction rather than overloading `dataSource`, so nothing downstream changes
  meaning. Happy to merge them if you prefer one field.
- **Spotless.** I have no JDK on this machine (JRE only), so I could not run
  `spotless:apply` locally. If CI flags formatting I will apply its output
  verbatim.

### DIFF
--- a/docs/process/equipment/separators.md
+++ b/docs/process/equipment/separators.md
@@ -1342,6 +1342,99 @@ per train — matching the format of the standard Sulzer
 
 ---
 
+## Constraint provenance: where a limit comes from
+
+Every `CapacityConstraint` carries two independent descriptors, and it is worth
+keeping them apart when reading a capacity report:
+
+| Descriptor | Question it answers | Type |
+|------------|---------------------|------|
+| `ConstraintSeverity` | How is a violation handled? | `CRITICAL`, `HARD`, `SOFT`, `ADVISORY` |
+| `ConstraintSource`   | What authority backs the number? | `CONFORMITY_STANDARD`, `VENDOR_DATASHEET`, `PROCESS_EMPIRICAL`, `USER_RULE`, `AUTO_SIZE`, `DEFAULT` |
+
+A vendor surge limit and a company K-factor may both be `HARD`, but they are not
+renegotiable on the same terms — the first is a machine limit, the second is a
+policy choice. Recording the source lets a report say *why* a limit binds.
+
+```java
+constraint.setSource(CapacityConstraint.ConstraintSource.CONFORMITY_STANDARD, "API 12J");
+constraint.getSource();           // CONFORMITY_STANDARD
+constraint.getSourceReference();  // "API 12J"
+```
+
+`ConstraintSource` is distinct from the free-text `dataSource`. The data source
+records *which part of the model* supplied the number (`"equipment"`,
+`"mechanicalDesign"`, `"installed_capacity_table"`); the constraint source
+classifies the underlying authority. Both are emitted per constraint in the
+`getUtilizationSnapshot()` JSON.
+
+### Hard, binding constraints from empirical observations
+
+The constraint that actually limits production on many field cases is not a
+standard's K-factor; it is an empirical relationship between an operating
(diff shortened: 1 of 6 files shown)


EvenSol (approved):
good update

PR #3451, opened 2026-09-04
Add pluggable entrainment models with a selectable carry-over default

dupuypablo:
## What

Separator carry-over becomes a **selectable model** rather than a single implicit path, and gains
a Service Provider Interface so a proprietary correlation can be supplied as a separate JAR
without any of its content entering this repository.

## Models shipped

| Id | Behaviour |
|----|-----------|
| `zero` | No carry-over. A deliberate baseline. |
| `spe-0.1gal-mmscf` | **Default.** Fixed 13.4 L of liquid per MSm³ of gas — the SI equivalent of the 0.1 US gal/MMscf rule of thumb from *The Savvy Separator: A Century of Carry-Over — 0.1 gal/MMscf* (SPE/JPT). Split oil/water at the feed water cut. |
| `neqsim-7stage` | The existing open-source physics chain, via `SeparatorPerformanceCalculator`. |

```java
separator.setEntrainmentProvider("neqsim-7stage");
double oil = separator.getEntrainmentResult().getOilInGasKgPerHr();
```

## Opt-in by construction

Selecting a model affects `getEntrainmentResult()` **only**. It does not change the entrainment
applied during `run()`, so adding this to an existing model cannot move that model's numbers.
`selectingAModelDoesNotChangeSeparatorResults` pins that promise rather than leaving it to
convention.

## Units — the part worth reviewing closely

- **`Sm3` in NeqSim is a molar quantity**: `n * R * T_std / atm`, no density. The gas basis is
  read directly as `MSm3/hr` from the gas outlet rather than derived from a density, which is the
  easy mistake here.
- **The 7-stage fractions are fractions of the incoming liquid**, not of the gas
  (`1 - overallGasLiquidEfficiency`). So the mass rate is `fraction × feed phase mass flow` — no
  density assumption at all.
- **Water cut** follows the definition already used elsewhere in NeqSim (`TwoFluidPipe`): water
  volume over total liquid volume.
- Conversion check for the default figure: 0.1 US gal = 0.37854 L against 1 MMscf = 28 262 Sm³
  (scf at 60 °F, Sm³ at 15 °C) = **13.39 L/MSm³**.

The spec model is verified by a **volume closure** test — reported masses are converted back to
volumes and compared against 13.4 L/MSm³ — so the test does not simply re-implement the model's
own arithmetic.

## The default is an assumption, not a prediction

`spe-0.1gal-mmscf` returns the same litres per MSm³ whatever the separator is doing — it does not
respond to gas load, internals or overload. That is the point of a rule-of-thumb default, but it
has one consequence I have stated on the class and in the docs rather than leaving to be
discovered: **a capacity constraint fed from this model can never trip on carry-over.**

## Notes for the reviewer

- `Separator` stores the selected model as a **`String` id, not a provider instance**, so the
  field stays serializable and needs no `transient` marker. (The original design held the
  provider object and needed one.)
- An unknown id fails loudly and names the models that *are* available, rather than silently
  substituting a different correlation.
- Providers declaring a higher SPI revision than the core supports are refused at lookup.
- `BuiltInSevenStageProvider` previously returned `NaN` on every channel. It now returns real
  mass rates. **The chain itself is not validated in this repository** — it is offered as one
  selectable model, not as a reference result. If its physics needs work, that is a separate
  change.
- I could not run `spotless:apply` locally (this machine has a JRE only, and the plugin needs
  Java 17). Formatting was applied with a rewrapper validated to reproduce Spotless's output
  byte-for-byte on an untouched file; if CI still flags anything I will apply its output verbatim.

## Follow-up, not in this PR

`docs/process/equipment/private-extensions.md` documents the implementer contract and the
public/private split, so an external party can write a provider without seeing anything private.

### DIFF
--- a/docs/process/equipment/private-extensions.md
+++ b/docs/process/equipment/private-extensions.md
@@ -1,0 +1,110 @@
+---
+title: "Private entrainment models"
+description: How to supply a proprietary carry-over correlation to NeqSim as a separate JAR, using the entrainment provider Service Provider Interface, without publishing the correlation itself.
+---
+
+NeqSim computes separator carry-over through a small Service Provider Interface (SPI), so a
+proprietary correlation can be supplied as a separate JAR without any of its content entering
+the public repository.
+
+## Models that ship with NeqSim
+
+| Id | What it does |
+|----|--------------|
+| `zero` | Reports no carry-over. Useful as a deliberate baseline. |
+| `spe-0.1gal-mmscf` | **Default.** A fixed 13.4 L of liquid per MSm³ of gas — the SI equivalent of the 0.1 US gal/MMscf rule of thumb from *The Savvy Separator: A Century of Carry-Over — 0.1 gal/MMscf* (SPE / JPT). Split between oil and water at the feed water cut. |
+| `neqsim-7stage` | The open-source 7-stage physics chain, evaluated through `SeparatorPerformanceCalculator`. |
+
+Select one on a separator:
+
+```java
+separator.setEntrainmentProvider("neqsim-7stage");
+EntrainmentResult result = separator.getEntrainmentResult();
+double oilCarryOver = result.getOilInGasKgPerHr();
+```
+
+With nothing selected the default applies. Selection affects `getEntrainmentResult()` only — it
+does not change the entrainment applied during `run()`, so adding it to an existing model cannot
+move that model's results.
+
+### The default is an assumption, not a prediction
+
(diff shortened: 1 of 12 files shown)


EvenSol:
CI refresh after merged #3452

- advanced the PR to tree-identical head `0a1d9948118672112b2483b3c3f442b85d217809` only to trigger a fresh merge-based check run against repaired `master`;
- verified merge commit `d04861f9d003f17aa38f9357a8a087673b3cee39` contains both `setAllowChoked(true)` selections;

PR #3452, opened 2026-09-04
[AI-Assisted] test(dynamics): preserve choked valve reference case

EvenSol:
## Summary

Restore the failing `ProcessSystemRunTransientTest.testDynamicCalculation1` reference case after #3448 made the legacy valve-sizing strategy unchoked by default.

The test's stored Cv anchors were generated on the choked IEC gas-sizing path. It now opts both gas valves into `setAllowChoked(true)` explicitly, preserving that established transient scenario while retaining #3448's new default and forward/inverse pressure consistency for ordinary callers.

## Root cause

#3448 intentionally changed the legacy `ControlValveSizing` choked-flow capacity limit from implicit-on to explicit-off. The dynamic test still expected its former choked Cv values, so the first SI Kv assertion changed from `1.1225433733413146` to `0.5133077841818874`.

This is a test-fixture configuration repair; no production behavior or public API changes.

## Validation

Local environment: OpenJDK 17.0.20, Maven 3.9.16.

- Reproduced the exact failure on current `master`.
- `./mvnw -q -Dtest=ProcessSystemRunTransientTest#testDynamicCalculation1 test` — passed after the repair.
- `./mvnw -q -Dtest=ProcessSystemRunTransientTest,ThrottlingValveTest test` — 32 tests passed, 0 failures/errors/skips.
- `python3 devtools/run_spotless.py apply` — passed.
- `python3 devtools/run_spotless.py check` — passed.
- `python3 devtools/check_documentation_search.py` — passed (688 Markdown + 1 standalone HTML page).
- `pre-commit run --all-files --hook-stage pre-commit` — passed.
- `pre-commit run --all-files --hook-stage pre-push` — passed.

Full repository matrix was not rerun locally; hosted CI remains authoritative for Java 8 and broad-suite validation.

## Documentation impact

Documentation impact: none. Only an existing test fixture now explicitly selects the already-documented choked-flow option; runtime behavior, API, units, defaults, and user guidance are unchanged.

PR #3455, opened 2026-09-04
fix(thermo): repair UNIFAC group tables and fail loudly on missing groups

dupuypablo:
## Problem

A component missing from a UNIFAC table, or carrying a wrong group assignment, does not fail. It ends up with `R = Q = 0`, which makes the combinatorial term

```
lngammaCombinational = log(V/x) + (10/2)*Q*log(F/V) + li - (V/x)*sum(l)
```

evaluate to NaN, because `V = x*R/sum = 0` and `F = x*Q/sum = 0`.

Observed effect, measured on a methane/ethane/third-component UMR-PRU fluid:

| third component | `calcPTphaseEnvelope` (Michelsen) |
|---|---|
| n-heptane (has groups) | 64 dew points, Tmax = 360.27 K |
| propylbenzene (no groups) | **1 point, Tmax = NaN** |

## Group parameters

Six R/Q values in `UNIFACGroupParam.csv` disagreed with the [DDBST published original UNIFAC table](https://www.ddbst.com/published-parameters-unifac.html), which this file already cites through its `Hansen1991` reference column.

| subgroup | was | DDBST | fingerprint |
|---|---|---|---|
| sub19 CH2CO | R 1.3357 | 1.4457 | typo |
| sub23 HCOO | Q 1.1898 | 1.1880 | digit transposition |
| sub44 CH2Cl | R 2.2564 | 1.4654 | had **CH2Cl2's** R |
| sub45 CHCl | R 2.0606 | 1.2380 | had **CHCl2's** R |
| sub58 CS2 | R 2.5070 | 2.0570 | digit transposition |

None of these subgroups is referenced by any component in the database, so the correction is **result neutral** for the shipped component set. Agreement goes from 59/65 to 65/65.

## Component tables

- **`oxygen` was assigned the N2 group** (`sub125`) in both tables, so oxygen was modelled as nitrogen. Now uses the O2 group (`sub123`).
- `cis-12-DM-cy-C6` and `trans-12-DM-cy-C6` carried a **water** group in a C8H16 naphthene.
- `E-cy-C5` was `CH3 + 6 CH2`, which is not a closed-shell molecule. Ethylcyclopentane is `CH3 + 5 CH2 + CH`.
- **Eleven rows were unreachable** because they used commas where `COMP.csv` uses dots, e.g. `"1,2,3-TM-Benzene"` against `1.2.3-TM-Benzene`. The lookup is `WHERE Name = '<COMP.csv name>'`, so those components silently got zero groups.
- **Four duplicate rows removed.** Which duplicate wins is otherwise decided by database row order. Each was resolved on molar mass and skeleton, never on position: `23-DM-C5` CompNumber 134 places a quaternary carbon in 2,3-dimethylpentane; the `c-C7` / `c-C8` duplicates were exact copies.

## MEG / DEG / TEG deliberately untouched

Left byte identical to master at the maintainer's request, since UMR-PRU is not generally used for glycol-bearing fluids.

Note the coupling: `sub61` (Furfural) is referenced by **exactly one** component in the database, MEG, so correcting its Q would in practice be a MEG change. That correction, the stray Furfural group on MEG, and the duplicate TEG row are therefore recorded in the baseline rather than fixed.

## Behaviour change

`ComponentGEUnifac` and `ComponentGEUnifacUMRPRU` now **throw** when a component ends up with no groups, wrapping `InvalidInputException` the way `SystemThermo.addComponent` already does. Previously the failure was swallowed by a `logger.error` and surfaced much later as NaN.

## New regression test

`UnifacDatabaseIntegrityTest` screens both component tables and the group table:

- **strict**: group R/Q and main group against DDBST, duplicate component names, unknown subgroups, rows with no groups
- **baselined** (53 entries): cross-table molar-mass consistency against `COMP.csv`, and rows with no `COMP.csv` component

A second test fails on stale baseline entries, so the baseline can only shrink.

Most of the remaining `molar_mass_mismatch` entries point at **`COMP.csv`**, not the UNIFAC tables. For `2-M-C8`, `3-M-C8`, `ethylcyclohexane`, `11-DM-cy-C5`, `113-TM-cy-C5` and `cis-13-DM-cy-C6` the group assignment is right and the `COMP.csv` molar mass is wrong.

## Verification

- `neqsim.thermo.**`: **1332 tests, 0 failures, 0 errors**
- `UnifacDatabaseIntegrityTest`: 2 tests, 0 failures
- **Negative control**: injecting a wrong CH3 R makes the test fail and name `group_param_mismatch sub1 CH3`
- `spotless:check` clean; data files verified UTF-8, no BOM, LF

## Not included

- 123 verified new component rows (from `devtools/propose_unifac_groups.py`) — adding components is a larger, separate review
- Convention migration: ~22 ring rows still use aliphatic `CH2` rather than the UMR-PRU cyclic `cCH2`, and 6 `nCx-Benzene` rows use `AC` rather than `ACCH2`. Both change results and need a before/after comparison first.


EvenSol:
CI refresh after merged #3452

- advanced the PR to tree-identical head `3baf8faa7e9861d4cbffc2878d7ee5aa0e3f9690` only to trigger a fresh merge-based check run against repaired `master`;
- verified merge commit `6b4ea20625820a3d3a63b05962ab40d527214d7e` contains both `setAllowChoked(true)` selections;

PR #3459, opened 2026-09-04
fix(thermo): stop addTBPfraction discarding the acentric factor it is given

dupuypablo:
## Problem

The seven-argument overload

```java
addTBPfraction(name, moles, molarMass, density, TC, PC, acentricFactor)
```

exists so a caller can supply **measured or externally regressed** values instead of correlations. It honoured `TC` and `PC` but **silently replaced the acentric factor**:

```java
acs = acentricFactor;                          // caller's value
...
if (characterization.getTBPModel().isCalcm()) {
  ...getAttractiveTerm().setm(m);              // m from calcm(molarMass, density)
  acs = ...getAcentricFactor();                // ← caller's value destroyed
}
```

`m` comes from a correlation. Pushing it onto the attractive term and reading the acentric factor back overwrites whatever was passed in. That same value was then read from the reference system when populating the real component, so the argument had **no effect at all**.

Supplying `0.49` gave `0.4796`.

This is the same class of defect as #3425 (`addTBPfraction2` discarding its boiling point): an overload that accepts a value and then ignores it.

## Fix

The correlation no longer overrides an explicitly supplied acentric factor. The attractive term derives its `m` parameter from the supplied value instead, which is self-consistent with the `TC` and `PC` that are already honoured.

**The four-argument overload is untouched** — deriving every property from correlations is its purpose.

## Verification

`AddTBPfractionAcentricFactorTest`, three cases:

1. the supplied acentric factor is retained
2. the supplied critical temperature and pressure are retained
3. two *different* acentric factors give two *different* stored values — so the test cannot pass by coincidence

**Negative control** — with the fix reverted, 2 of 3 fail with exactly the reported symptom:

```
expected: <0.49> but was: <0.47958681250883073>
```

**Regression** — `neqsim.thermo.**`, `neqsim.pvtsimulation.**`, `neqsim.thermodynamicoperations.**`: **2273 tests, 0 failures, 0 errors**.

## Documentation

`docs/cookbook/thermodynamics-recipes.md` gains an overload-selection table stating the unit of every argument and which values are honoured as given versus correlated:

| Overload | Supply | Derived from correlations |
|----------|--------|---------------------------|
| 4-arg | molar mass, density | TC, PC, acentric factor, boiling point |
| 7-arg | molar mass, density, TC, PC, acentric factor | boiling point |
| `addTBPfraction2` | molar mass, boiling point | density, TC, PC, acentric factor |
| `addTBPfraction3` | density, boiling point | molar mass, TC, PC, acentric factor |

## Note on behaviour change

Callers of the 7-argument overload will now get the acentric factor they asked for. Results will move for anyone who was passing a value that differed from the `calcm` correlation — which is the point of the fix.

### DIFF
--- a/docs/cookbook/thermodynamics-recipes.md
+++ b/docs/cookbook/thermodynamics-recipes.md
@@ -67,6 +67,24 @@ by `characterisePlusFraction()` when a residual fraction is to be split and lump
 characterization parser does not safely parse that name. Use a numeric terminal carbon label such
 as `C7` when calling `characterisePlusFraction()`.
 
+#### Choosing an `addTBPfraction` overload
+
+| Overload | Supply | Derived from correlations |
+|----------|--------|---------------------------|
+| `addTBPfraction(name, moles, molarMass, density)` | molar mass, density | TC, PC, acentric factor, boiling point |
+| `addTBPfraction(name, moles, molarMass, density, TC, PC, acentricFactor)` | molar mass, density, TC, PC, acentric factor | boiling point |
+| `addTBPfraction2(name, moles, molarMass, boilingPoint)` | molar mass, boiling point | density, TC, PC, acentric factor |
+| `addTBPfraction3(name, moles, density, boilingPoint)` | density, boiling point | molar mass, TC, PC, acentric factor |
+
+Units: `molarMass` in kg/mol, `density` as relative density (g/cm3), `TC` in K, `PC` in bara,
+`boilingPoint` in K, `acentricFactor` dimensionless.
+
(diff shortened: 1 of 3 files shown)


EvenSol:
CI refresh after merged #3452

- advanced the PR to tree-identical head `bfbb10b4a56e355a8c36601b9f57a1f978a09238` only to trigger a fresh merge-based check run against repaired `master`;
- verified merge commit `90289ba9432acabc70e60d9b90aa516284c31792` contains both `setAllowChoked(true)` selections;

PR #3460, opened 2026-09-04
[publish-release] release: NeqSim 3.20.0

dupuypablo:
Bumps the active version from **3.19.0 to 3.20.0**.

## Change

17 version replacements across the same 12 files used by the 3.19.0 release (#3438): the Maven coordinates, `CITATION.cff`, and the quickstart / wiki / README snippets that pin a version for users. The PR also updates the release workflow, for 13 changed files in total.

The replacement is restricted to an **explicit file list** rather than a repository-wide search and replace. `commons-lang3` also passed through 3.19.0; it currently sits at 3.20.0 on its own dependency line and is untouched.

## Verification

- `mvn compile` reports `Building NeqSim 3.20.0`, **BUILD SUCCESS**
- No NeqSim `3.19.0` reference remains (`<revision>`, `version:`, or artifact coordinate)
- `CITATION.cff` `date-released` already reads `2026-09-08`

## Contents of this release

Component and UNIFAC database integrity, plus two overloads that silently discarded arguments they were given:

| PR | |
|---|---|
| #3425 | `addTBPfraction2` discarded its boiling point |
| #3426 | component synonyms were not resolved in `Phase.getComponent` |
| #3455 | UNIFAC group tables repaired; a component with no groups now fails loudly instead of producing NaN |
| #3457 | `COMP.csv` component identities repaired, integrity gate added |
| #3458 | `COMP.csv` rows holding another substance's properties corrected |
| #3459 | `addTBPfraction` discarded the acentric factor it was given |

**Merge after those have landed**, so the release contains them.

## AI-assisted CI repair — September 8

The failed [pre-commit run](https://github.com/equinor/neqsim/actions/runs/34207745142) checked an older merge result containing an unformatted `UMRPRUGroupTableFlashQualificationTest.java` from master. PR #3570 already applied the hosted Spotless formatting to that test in `3db4ec373a3612ae069abf6ed068969b3db7514e`.

Commit `ab67a029d06ae7939abe0aef90a37a06fe7ec05e` refreshes CI with exactly the same release tree as `7f0bd8126b969c65c1b1c357f9300a225b1a1029`. GitHub's new test merge `e8c91c69158aa5b20e8140b8a7e9b3696c197249` has the same tree as the locally validated merge with master `3db4ec373a`. The 13-file PR diff is unchanged.

Local validation on that prospective merge, using the provided primary-runtime Python and cached pre-commit dependencies after Maven bootstrap with `prepare_maven.py`:

[...]

PR #3549, opened 2026-09-07
Complete and correct the UMR-PRU UNIFAC group table

dupuypablo:
## Why

UMR-PRU is the model in active use, so `UNIFACcompUMRPRU.csv` is the table that
has to be complete and correct. `UNIFACcomp.csv` serves the original UNIFAC and
UNIQUAC models and is left **exactly as published** — this PR does not touch it.

A component missing from a UNIFAC table does not fail cleanly. It gets R = Q = 0,
the combinatorial term goes NaN, and a phase envelope quietly collapses to a
single point.

## The coupling that caused the drift

`PhaseGEUnifac`'s constructor built classic `ComponentGEUnifac` objects, which
`PhaseGEUnifacUMRPRU` and `PhaseGEUnifacPSRK` then **discarded and rebuilt** from
their own table. Because the classic constructor reads `UNIFACcomp.csv`, every
UMR-PRU component was silently required to have a row in a table it never uses.

A fluid built explicitly with `setMixingRule("HV", "UNIFAC_UMRPRU")` would fail
with *"no UNIFAC group assignment in **UNIFACcomp.csv**"*.

`PhaseGEUnifac` now returns early when it is constructing a subclass, matching
the idiom already used in `ComponentGEUnifac`.

This is what had been pulling hydrocarbon rows into the classic table.

## Rings use the cyclic groups

27 naphthene rows described ring carbons with aliphatic CH2/CH. Four were
internally inconsistent within a single molecule — `M-cy-C5` had a cyclic ring
CH2 next to an acyclic ring CH. Others were fully acyclic, including `c-C7` as
seven plain CH2.

They now use `cCH2` (136) and `cCH` (137), which matches the **DDBST modified
UNIFAC** assignment set (its subgroups 78/79/80 are `CY-CH2`, `CY-CH`, `CY-C`):

| molecule | DDBST modified UNIFAC |
|---|---|
| cyclohexane | `{CY-CH2: 6}` |
| methylcyclohexane | `{CH3: 1, CY-CH2: 5, CY-CH: 1}` |
| n-butylcyclohexane | `{CH3: 1, CH2: 3, CY-CH2: 5, CY-CH: 1}` |
| cyclopropane | `{CY-CH2: 3}` |

`cCH2` carries the **same R and Q** as `CH2`, so the combinatorial term is
untouched and only the residual term moves. The rows are equal except against
eight main groups: water, CO2, CH4, N2, H2S, C2H6, Hg and TEG.

The rewrite was driven off the group assignment rather than the `FORMULA`
column, which is not reliable, and refuses to write if a rewrite changes the
carbon or hydrogen count.

## Every hydrocarbon now resolves

12 rows added, each checked against structure and against DDBST where DDBST has
an entry:

```
c-propane           3 cCH2
c-C4                4 cCH2
n-Bcychexane        CH3 + 3 CH2 + 5 cCH2 + cCH
Pent-CC6            CH3 + 4 CH2 + 5 cCH2 + cCH
cis-14-DM-cy-C6     2 CH3 + 4 cCH2 + 2 cCH
trans-14-DM-cy-C6   same; UNIFAC has no stereochemistry
argon               Ar 124
hydrogen            H2 127
ortho-hydrogen      H2 127, spin isomer of the same molecule
para-hydrogen       H2 127
propanePVTsim       alias of propane
nbutanePVTsim       alias of n-butane
```

`everyHydrocarbonHasAnUmrPruGroupAssignment` enforces this from now on. Unlike
the baselined findings it has **no room to grow**: a new hydrocarbon has to
arrive together with its group assignment. Verified to fail by renaming a row:

```
Hydrocarbons missing from UNIFACcompUMRPRU.csv: [argon].
Add the group assignment in devtools/add_unifac_component_rows.py
rather than skipping the component.
```

## The one hydrocarbon left out

**Ethylene** is allowlisted, with the reason recorded in the test. Main group 2
offers only substituted olefin subgroups — CH2=CH, CH=CH, CH2=C, CH=C, C=C — and
none of them stands for a bare CH2=CH2. DDBST has no assignment for it in its
original, modified or PSRK sets either.

Covering ethylene needs a **regressed group of its own**, the way `Voutsas`
added C2H6 as group 134. That is a modelling contribution, not a data-entry fix,
and it must not be approximated with a substituted olefin group.

## Baseline

`missing_unifac_row` is now recorded for the UMR-PRU table only. The 81 entries
against the classic table were noise: that table is frozen, so a component absent
from it is not a gap to be filled.

Baseline drops **189 → 95**.

## Test expectation changes

Seven pinned numbers move, across three test classes. All three fluids contain
`c-hexane`, `c-C7` and `c-C8`, and all three previously described `c-hexane` as
a ring while calling the other two chains.

**Attribution was measured, not assumed.** Restoring only the master version of
`UNIFACcompUMRPRU.csv`, with every Java change in this branch still in place,
makes all eight tests pass again — so the `PhaseGEUnifac` change is numerically
inert and the ring conversion is the sole cause.

`SystemUMRPRUMCEosNewTest`, cricondenbar:

| | before | after |
|---|---|---|
| T | −11.09948347 °C | −11.492052658160333 °C |
| P | 104.75329137038476 bara | 104.69878072485776 bara |

`SaturationTemperatureTest`:

| | before | after |
|---|---|---|
| cricondentherm T | 23.469 °C | 23.42135124839797 °C |
| cricondentherm P | 46.9326702068279 bara | 47.400609467549636 bara |
| sat T at cricondentherm | 23.469396812206867 °C | 23.421359634399437 °C |
| issue 3269 sat T | 29.4644 °C | 29.430432891845726 °C |

`UMRPRUOilDropoutReproTest`, oil dropout:

| | before | after | Δ |
|---|---|---|---|
| 8 °C | 9.144142e−04 | 8.910901e−04 | −2.6 % |
| 18 °C | 3.502455e−04 | 3.374951e−04 | −3.6 % |
| 20 °C | 2.582079e−04 | 2.477208e−04 | −4.1 % |

Every quantity moves in the **same direction**: the naphthenes become slightly
more compatible with the gas, so dew temperatures fall and less liquid drops out.

None of these numbers has an external basis. `UMRPRUOilDropoutReproTest` says so
itself — its javadoc describes the values as *"deterministic numerical evidence,
not experimental PVT validation"* and as a baseline established by an earlier
repair.

**No physics-based assertion was touched, and all still hold**: material balance
to 1e−10, fugacity equality to 1e−8, monotonic dropout with falling temperature,
and the closure that the saturation temperature at the cricondentherm pressure
equals the cricondentherm temperature — now to 8e−6 °C.

## Documentation

`docs/thermo/component_database_guide.md` now states which table each model
reads, that UMR-PRU is the maintained one, that NTUA/Voutsas takes precedence
over DDBST, that `Secondary` (not `ID`) is the subgroup number, and why the same
molecule legitimately has different assignments in the two tables.

The absence of that note is the root cause of the drift this PR cleans up.

## Verification

Full suite green.


dupuypablo:
Targeted at **3.20.0**. The release PR #3460 is held as a draft until this
merges, so this is on the critical path for the release.

Reviewing notes, shortest path first:

- The behaviour change is confined to eight main-group interactions
  (H2O, CO2, CH4, N2, H2S, C2H6, Hg, TEG). `cCH2` carries the same R and Q as
  `CH2`, so the combinatorial term is untouched.
- Seven pinned test numbers move. Attribution was measured, not assumed:
  restoring only master's `UNIFACcompUMRPRU.csv`, with every Java change here
  still in place, makes all the affected tests pass again.
- No physics-based assertion was changed. Material balance (1e-10), fugacity
  equality (1e-8), monotonic dropout and the cricondentherm closure all still
  hold on the original tolerances.
- Full suite run locally before rebase: 13043 tests, 0 failures.

PR #3725, opened 2026-09-15
Add 132 components to COMP.csv and 131 UMR-PRU rows, correct 228 data defects

dupuypablo:
## What this does

Extends the component database so gas-chromatography slates containing branched paraffins, alkylbenzenes, cycloalkanes and light olefins can be modelled with **UMR-PRU**, and corrects data defects found while auditing the existing rows.

| file | before | after |
|---|---|---|
| `COMP.csv` | 257 rows, 170 columns | **389 rows, 171 columns** |
| `UNIFACcompUMRPRU.csv` | 146 rows | **277 rows** |

No row present before is removed, no column is removed, no duplicate names are introduced.

> The diff looks larger than the change because adding the `InChIKey` column rewrites every line: 258 deleted + 390 inserted = 648 for `COMP.csv`, 147 + 278 = 425 for the UNIFAC table.

## New `InChIKey` column

A canonical structure identifier for every row that resolves (366 of 389). Unlike a CAS number it is *computed from the structure*, so it is one value per substance regardless of naming, and it is how the DDBST UNIFAC tables are indexed.

It paid for itself immediately — it surfaced that the **`glycerol` row is a byte-for-byte copy of `TEG`** (same CAS 112-27-6, same C6H14O4, same 150.175). Real glycerol is 56-81-5 / C3H8O3 / 92.094. Anyone calling `addComponent("glycerol", ...)` silently got triethylene glycol.

## Data corrections — 228 across 154 rows

Every value was checked against **two independent sources**: the UniSim Design traditional component library (2178 components, joined on CAS) **and** the median of the published *measured* sources compiled in `chemicals` (IUPAC, CRC, PSRK, Poling-Danner, NIST WebBook, Pina-Martines, Yaws). A value was changed only where it disagreed with **both**. Estimation methods (Joback, Wilson-Jasperson, Fedors) were excluded from the consensus.

| column | corrections |
|---|---|
| `LIQDENS` | 141 |
| `NORMBOIL` | 46 |
| `MOLARMASS` | 21 |
| `FORMULA` | 12 |
| `CASnumber` | 8 |

Representative defects:

- **`11-DM-cy-C5`, `113-TM-cy-C5`, `cis-13-DM-cy-C6`** each carried `NORMBOIL = 58.05 degC` — isopentane's boiling point, and the modal value in the whole file — against true values of 87.8, 104.9 and 124.4 degC. Their `FORMULA` said `C7H16` where two of them are `C8H16`, and their molar mass was the *acyclic* isomer's, exactly 2.015 g/mol too high because ring closure was never applied.
- **`methylcyclohexane`** carried `LIQDENS = 0.4222` — methane's liquid density at its normal boiling point — against a true 0.7739.

**`TC`, `PC` and `ACSFACT` were deliberately not changed.** They may be deliberately tuned, so the 84 deviations found there are reported for review rather than corrected.

Two of my own early findings were **retracted** during this work: `ethylcyclohexane` and the Tc/Pc of `cis-1,3-dimethylcyclohexane` are correct in `COMP.csv` — the apparent deviation came from comparing against a single preferred source where CRC happens to be the outlier. They agree with four or five other published sources.

The `LIQDENS` convention was established from the data rather than assumed: density at 15 degC for a component that is liquid there, density at the normal boiling point otherwise (methane 0.4222 = rho@NBP 0.4224; toluene 0.8710 = rho@15C 0.8716). UniSim's `StdLiquidDensity` is a *different quantity* for light gases (0.2994 for methane) and was therefore **not** used as a source for this column.

## UMR-PRU group assignments

Generated from the DDBST published **modified**-UNIFAC assignments, resolved by InChI key, with DDBST's cyclic subgroups 78/79/80 remapped onto NeqSim's 136/137/138 (`cCH2`/`cCH`/`cC`).

Every candidate was gated on an **independent physical check**: the molar mass implied by the summed subgroup formulas must match `MOLARMASS`. Anything failing a step was rejected, never approximated — an unverified assignment produces a silent `NaN` rather than an error.

| origin | rows |
|---|---|
| direct DDBST lookup | 126 |
| via the stereo-generic InChI key | 4 |
| hand-derived from the IUPAC name | 1 |

The stereo-generic fallback is **exact, not an approximation**: UNIFAC counts groups and cannot distinguish stereoisomers. `trans-1-ethyl-2-methylcyclopentane` (`BSKOLJVTLRLTHE-HTQZYQBOSA-N`) and the stereo-undefined form (`BSKOLJVTLRLTHE-UHFFFAOYSA-N`) return identical DDBST assignments.

**No new groups are introduced**, so no interaction parameter is required — `UNIFACGroupParam.csv` and the `UNIFACInterParam*` tables are untouched. The three subgroups used here for the first time — 8 (`CH=C`), 13 (`ACCH`), 70 (`C=C`) — belong to main groups 2 and 4, which were already in use.

## :warning: Known limitation — ethylene (TODO)

**`ethylene`, `C2H4` and `C2H4-` still have no UMR-PRU row and cannot be used with that model.**

The UMR group set has `C2H6` (main group 64) for ethane but **no C2-olefin analogue**, and main group 2 offers only *substituted* olefins (`CH2=CH`, `CH=CH`, `CH2=C`, `CH=C`, `C=C`) — none of which is a bare CH2=CH2. DDBST has no entry for it in any of its three published sets. This is confirmed against Table 1 of the UMR-NEW model validation report, whose complete group set is CH4, C2H6, CO2, N2, H2S, H2O, Hg, CH2, cCH2, ACH, ACCH2/ACCH3.

Resolving it needs a **new regressed group plus a full row and column of interaction parameters** — a parameter regression, not a data edit. Deliberately left out of scope.

Also still without a UMR-PRU row, for the same class of reason:

- **`5-methyl-3-heptyne`** — needs the alkyne group, which is absent from `UNIFACGroupParam.csv`
- **`asphaltene`** — a pseudo-component with no structure
- **`4-ethyl-3-heptene`, `5-ethyl-3-methylheptane`** — no resolvable CAS and no source data, so they are **deliberately not added** rather than added with invented properties

## Verification

Loaded into the engine and exercised end to end:

```
   COMP                 389 rows
   UNIFACcompUMRPRU     277 rows

--- existing component unchanged ---
  PASS  methane baseline: MW=16.043  Tc=-82.59 degC

--- newly added components ---
  PASS  '1,2,4-trimethylbenzene': MW=120.192  Tc=375.95 degC  Tb=169.40 degC  rho_l=0.8797
  PASS  'cumene': MW=120.192  Tc=357.85 degC  Tb=152.40 degC  rho_l=0.8648
  PASS  '2,4-dimethylheptane': MW=128.255  Tc=303.65 degC  Tb=132.00 degC  rho_l=0.7179
  PASS  '1,1,2-trimethylcyclopentane': MW=112.213  Tc=306.39 degC  Tb=113.70 degC  rho_l=0.7876
  PASS  'trans-1,3-dimethylcyclohexane': MW=112.213  Tc=324.85 degC  Tb=124.46 degC  rho_l=0.7163
  PASS  '1-ethyl-2-methylcyclopentane', '3-methylnonane', 'mesitylene',
        'isobutylbenzene', '1,3-diethylcyclopentane'

--- UMR-PRU (a missing UNIFAC row surfaces as NaN) ---
  PASS  UMR-PRU aromatic:          2 phase(s), density=67.381 kg/m3
  PASS  UMR-PRU ring:              2 phase(s), density=66.175 kg/m3
  PASS  UMR-PRU branched paraffin: 2 phase(s), density=89.647 kg/m3
  PASS  UMR-PRU olefin:            2 phase(s), density=69.787 kg/m3

SMOKE TEST: PASS
```

The UMR-PRU checks carry an explicit NaN guard on density, which is how a missing UNIFAC row manifests.

## Related follow-up (not in this PR)

A missing group-interaction parameter is currently **swallowed**: `PhaseGEUnifacUMRPRU.calcaij()` catches the exception from the empty result set, logs it, and leaves `aij[i][j]` at its default `0.0` — an *ideal* interaction. The run completes and returns plausible-looking numbers. This already affects the 36 cycloalkanes in the table when the non-Mathias-Copeman attractive term is selected, since `UNIFACInterParamA_UMR.csv` stops at main group 64 and has no cell for `cCH2`. It should throw instead. Raised separately.

## CI regression repair — 2026-09-15

Continued from exact head `326e12388b1f4ada5bc9b2d7b0eea62d6df7d350`; new head `c8747e30321d4170a3dffd19a4752442224fa102`.

The Windows matrix exposed 18 failures and one error. The extra `InChIKey` column made the positional temporary-component INSERT incompatible with the table, dropping TBP/wax metadata. Named-column inserts now preserve every physical parameter and tolerate optional identity columns. No wax baseline was relaxed.

The canonical resolver now includes all 389 standard names. CSV tests use the existing H2 CSV parser, so names containing commas are tested intact. Nine full standard names were missing from the independently maintained extended resource; loading extended mode now imports missing standard names with fresh IDs and optional columns while preserving existing extended properties. This keeps future standard additions available without rewriting the 86 MB extended resource.

Validation: original failures reproduced; **106 focused tests passed, zero failures/errors/skips**, including all failing wax, oil-quality, lookup and database suites plus UNIFAC integrity and duplicate-data guards. Direct Spotless apply (twice), Spotless check, documentation-search audit and `git diff --check` passed (exit 0). Local pre-commit executable unavailable; direct equivalents ran. The component database guide documents optional metadata and extended-mode behavior.

**VALIDATION PENDING UPDATED CI.** No merge, rebase or branch synchronization was performed.


dupuypablo:
Follow-ups split out of this PR so they can be decided independently:

- #3726 — **ethylene has no UMR-PRU group assignment.** The known limitation noted above. It needs a regressed `C2H4` group plus a full row and column of interaction parameters, so it is a parameter regression rather than a data edit.
- #3727 — **a missing UNIFAC interaction parameter is silently treated as zero.** Pre-existing, not introduced here, but worth deciding on: it is reachable today by 36 of the 146 original rows and by 24 of the rows added here, and it degrades results without failing.

Direct links to the two files as they stand in this PR, pinned to the commit so they survive the branch being deleted on merge:

- [`COMP.csv`](https://github.com/equinor/neqsim/blob/635e204119f0f53755b49f3307ddaa2b1c0e8875/src/main/resources/data/COMP.csv) · [raw](https://raw.githubusercontent.com/equinor/neqsim/635e204119f0f53755b49f3307ddaa2b1c0e8875/src/main/resources/data/COMP.csv)
- [`UNIFACcompUMRPRU.csv`](https://github.com/equinor/neqsim/blob/635e204119f0f53755b49f3307ddaa2b1c0e8875/src/main/resources/data/UNIFACcompUMRPRU.csv) · [raw](https://raw.githubusercontent.com/equinor/neqsim/635e204119f0f53755b49f3307ddaa2b1c0e8875/src/main/resources/data/UNIFACcompUMRPRU.csv)

dupuypablo:
Note for review: the `Detect Java/XML changes` gate correctly finds no Java or XML in this PR, so the whole test matrix is skipped:

```
Assert tests with Java 8 on Ubuntu    SKIPPED
Assert tests with Java 8 on Windows   SKIPPED
Assert tests with Java 21 on Ubuntu   SKIPPED
Assert slow tests                     SKIPPED
```

That gate is right for source changes, but these two CSVs are loaded at runtime by `NeqSimDataBase.updateTable()` and feed every flash, so **CI is not exercising them here**. A malformed row would pass green.

That is why the verification in the description was run locally against the built classes rather than relied on from CI: 389 and 277 rows load, `methane` is bit-for-bit unchanged, ten of the newly added components flash, and four UMR-PRU mixtures covering an aromatic, a ring, a branched paraffin and an olefin return finite densities behind an explicit NaN guard.

Worth considering separately whether the detect-changes gate should also trigger on `src/main/resources/data/**`.

dupuypablo:
Two follow-ups pushed in `e5a6456d`, plus a full cross-file consistency audit.

## Six names put into normal chemical order

Six components added here carried the CAS-index (inverted) form the source data used:

```
benzene, 1-butyl-2-methyl-        ->  1-butyl-2-methylbenzene
benzene, 2-ethyl-1,3-dimethyl-    ->  2-ethyl-1,3-dimethylbenzene
cyclohexene, 1-methyl-            ->  1-methylcyclohexene
cyclopentane, 1,2,3-trimethyl-    ->  1,2,3-trimethylcyclopentane
cyclopentane, 1-ethyl-3-methyl-   ->  1-ethyl-3-methylcyclopentane
octane, 2,7-dimethyl-             ->  2,7-dimethyloctane
```

The `NAME` column is the literal string passed to `addComponent`. `ComponentNameResolver` *does* de-invert via `deinvertCasIndexName()`, but it is **not referenced anywhere in `SystemThermo`** — only `ComponentInterface` exposes it — and its `CANONICAL_NAMES` map is hardcoded Java containing none of the components added here. So the raw name is what a user actually has to type.

None of the six exist on master, so nothing that currently works is renamed, and no de-inverted form collides with an existing name.

Both files are renamed together, because the engine joins them on the name string:

```java
SELECT * FROM unifaccompumrpru WHERE Name='" + name + "'
```

Renaming only one side would leave the component without a group assignment. Verified on the built classes: all six load under the new name with unchanged properties, none of the old names resolves any more, and five UMR-PRU mixtures built on the renamed components return finite densities.

## Cross-file consistency audit

Checked `COMP.csv`, `UNIFACcompUMRPRU.csv`, `UNIFACcomp.csv`, `COMP_EXT.csv` and `UNIFACGroupParam.csv`, reported as a **delta against `origin/master`** so pre-existing conditions are not attributed to this branch:

```
INTRODUCED BY THIS BRANCH (0)
  (none)

PRE-EXISTING, STILL PRESENT (4)
  ~ COMP.csv: hydrocarbon 'default' has no UMR-PRU row
  ~ UNIFACcompUMRPRU.csv: orphan row '3-E-C5' matches no COMP.csv NAME
  ~ UNIFACcompUMRPRU.csv: orphan row 'c-C8de' matches no COMP.csv NAME
  ~ UNIFACcompUMRPRU.csv: orphan row 'iC5'    matches no COMP.csv NAME
```

Covered: duplicate `NAME`/`ID`/`CompNumber`, ragged rows, orphan UNIFAC rows, hydrocarbons without a group assignment, and subgroups referenced without R/Q parameters in `UNIFACGroupParam.csv`.

The three orphans are harmless legacy rows left from older naming — `i-pentane` and `3-ethylpentane` each already have their own working UMR-PRU row, and `c-C8de` has no component at all. They are pre-existing and untouched here.

One check worth keeping: `UNIFACGroupParam.csv` keys subgroups on the **`Secondary`** column, not the first field. Reading the wrong column makes subgroups 121-127 and 134-138 look unparameterised when they are present.

## Note on the branch

`master` was merged into this branch remotely while the rename was being prepared. I verified the merge left both CSV blobs byte-identical to `635e2041` before rebasing on top of it — no force-push, nothing overwritten.

dupuypablo:
## Correction to my earlier comment

**I was wrong when I said `ComponentNameResolver` is "not wired into `SystemThermo.addComponent`".** It is. `addComponent` calls `ComponentInterface.getComponentNameFromAlias` (`SystemThermo` lines 354, 407, 438, 478, 507), which delegates to `ComponentNameResolver.resolve`. My grep looked for the literal class name inside `SystemThermo.java`, where it does not appear. The documentation was right.

The rename still stands, for a narrower reason: `resolve()` returns the input unchanged when nothing matches, and `CANONICAL_NAMES` is a hardcoded Java map that contains none of the components added here. So for these, the raw `NAME` is still exactly what has to be typed.

## The integrity gates were failing

`ComponentDatabaseIntegrityTest` and `UnifacDatabaseIntegrityTest` never ran on this branch — `Detect Java/XML changes` skipped the whole matrix because the change was CSV-only. Run locally, **all five tests failed**. Now:

```
ComponentDatabaseIntegrityTest   Tests run: 2, Failures: 0
UnifacDatabaseIntegrityTest      Tests run: 3, Failures: 0
BUILD SUCCESS
```

### Five density edits were genuinely wrong

`implausible_watson_k` fired on nC27, nC29, nC30, nC34, nC39 — and it was right.

The audit rule was "liquid density at 15 °C if liquid there, at the normal boiling point otherwise". These are **solid at 15 °C**, so the fallback took a density at 420–520 °C where the liquid is strongly expanded:

| | master | this branch | Watson K before | after |
|---|---|---|---|---|
| nC30 | 0.871 | 0.6831 | 12.52 | **15.97** |
| nC39 | 0.871 | 0.7383 | 9.65 | **15.23** |
| nC34 | 0.871 | 0.7450 | 12.71 | **14.85** |
| nC27 | 0.871 | 0.7657 | 12.36 | **14.06** |
| nC29 | 0.871 | 0.7773 | 12.41 | **13.97** |

Watson K is ~12.5–13.5 for an n-paraffin, so those cannot be right. **Reverted.** Master's 0.871 is itself a placeholder shared by 37 rows rather than a measurement, but it is at least consistent with the boiling point, and I do not have a better value I can defend — so reverting, not substituting, is the honest outcome.

The other 58 LIQDENS edits stand, and 9 of them repair values that were implausible before. The nC39 `NORMBOIL` fix (58.05 → 518.76 °C) is kept: 58.05 was isopentane's boiling point and gave a Watson K of 9.65 for a C39 paraffin.

### `5-methyl-3-heptyne`

Added to `HYDROCARBONS_WITHOUT_A_GROUP` beside `ethylene`. DDBST assigns it alkyne subgroup 66, which has no row in `UNIFACGroupParam.csv` and so has neither R and Q nor interaction parameters. Assigning an olefin group instead would silently give wrong activity coefficients.

### Baselines

```
comp_known_issues.tsv    121 -> 112   (-32 fixed, +23 new)
unifac_known_issues.tsv   95 ->  58   (-38 fixed,  +1 new)
```

The **70 removed entries are defects this branch actually repairs** — the formula/mass mismatches on the trimethylcyclopentanes, the Watson K failures from rows carrying methane's liquid density, the shared 58.05 °C boiling point. That is independent confirmation of the correction work.

The 23 added are `over_shared_formula` / `over_shared_molarmass`. Those checks look for values shared by more components than coincidence explains, to catch rows created by copying a template; adding many structural isomers legitimately shares formula and molar mass.

> :warning: **The baselines were edited from the test output, not regenerated.** The checks are implemented twice — in Python under `devtools/` and again in Java in the tests — and they do not agree. `screen_unifac_tables.py` emits **no `missing_unifac_row` finding at all**, so following the documented `--tsv >` regeneration drops 85 entries the tests still require and makes the failure worse. Both of these are now written down in the database guide.

## Documentation

- `component_list.md` — count 257 → **389**; the by-category table replaced with values counted from `COMPTYPE` instead of the previous approximations (`~50`, `~15`, `~150+`); the UMR-PRU table and its exact-name join documented; `InChIKey` introduced; and the components that cannot be used with UMR-PRU called out.
- `component_database_guide.md` — `InChIKey` added to the column reference with what it is actually good for; the ethylene section extended to cover alkynes; and two warnings recorded: that the screeners and tests disagree, and that **CI does not run these gates for a data-only change**, with the local command (note the comma — `-Dtest=A+B` matches nothing).

`spotless:apply` run on the edited test class.

EvenSol (approved):
Great update!

Issue #3739, opened 2026-09-16
calcPTphaseEnvelope does not converge on UMR-PRU for ordinary natural gas, and returns unconverged points


dupuypablo:
`calcPTphaseEnvelope()` does not converge on `SystemUMRPRUMCEosNew` for ordinary natural-gas mixtures. It returns an empty result, a single point, or physically impossible values, while SRK traces a clean envelope for the identical composition.

This is not caused by any particular component — a dry methane/ethane/propane gas on an untouched database fails.

## Reproduction

```python
from neqsim import jneqsim

Ops = jneqsim.thermodynamicoperations.ThermodynamicOperations
MIX = [("methane", 0.90), ("ethane", 0.07), ("propane", 0.03)]

for EOS, rule in ((jneqsim.thermo.system.SystemUMRPRUMCEosNew, ("HV", "UNIFAC_UMRPRU")),
                  (jneqsim.thermo.system.SystemSrkEos, ("classic",))):
    f = EOS(288.15, 50.0)
    for n, x in MIX:
        f.addComponent(n, x)
    f.setMixingRule(*rule)
    ops = Ops(f)
    ops.calcPTphaseEnvelope()
    env = ops.getOperation()
    print(EOS, len(env.getDewPointTemperatures()), len(env.getBubblePointTemperatures()))
```

Observed, on an unmodified install (`COMP` = 257 rows, nothing overridden):

```
UMR-PRU   dry gas envelope:  0 usable points
SRK       dry gas envelope: 63 usable points (33 bubble / 30 dew)
```

"Usable" means finite, `T > 1 K`, `P > 1e-6 bara`; the arrays are otherwise padded with NaN and zeros.

## Failure modes seen

Across a range of gas compositions, UMR-PRU produced all of the following:

| composition | result |
|---|---|
| methane/ethane/propane | empty |
| gas + 1% n-heptane | empty |
| gas + 1% cumene | empty |
| gas + 5% cumene | **1 point** |
| gas + 0.5% cumene | 422 points spanning **T = 38 915 to 76 527 °C** |
| gas + 1% toluene | 9979 points, but a **single branch** (0 bubble / 9979 dew) |

The 38 915 °C case is the most useful diagnostic: the continuation is not merely
failing to start, it is marching off into a non-physical region and returning
those points as if they were converged.

Two things make this easy to miss:

- the arrays are returned padded, so a caller that does not filter for finite,
  positive values sees a populated result;
[...]

PR #3753, opened 2026-09-17
fix(thermo): remove the 100-component cap in the Michelsen phase envelope

dupuypablo:
Fixes a hard 100-component limit in the Michelsen phase envelope, reported by a user whose characterised fluid has 101 components.

## The bug

`PTPhaseEnvelopeMichelsen` stores the phase composition at the cricondentherm and cricondenbar in arrays declared with a fixed length:

```java
private double[] cricondenThermX = new double[100];
private double[] cricondenThermY = new double[100];
private double[] cricondenBarX   = new double[100];
private double[] cricondenBarY   = new double[100];
```

but fills them with loops bounded by the component count:

```java
for (int ii = 0; ii < nonLinSolver.numberOfComponents; ii++) {
  cricondenThermX[ii] = system.getPhase(1).getComponent(ii).getx();
  cricondenThermY[ii] = system.getPhase(0).getComponent(ii).getx();
}
```

Any fluid with more than 100 components fails as soon as the trace reaches the first cricondentherm or cricondenbar update:

```
java.lang.ArrayIndexOutOfBoundsException: Index 100 out of bounds for length 100
```

The failure is not at `addComponent` and not at the start of the envelope, which is what makes it confusing to diagnose: the calculation begins normally and dies partway through the continuation.

## The fix

The arrays hold one mole fraction per component, so their length has to follow the fluid. They are now allocated from `system.getPhase(0).getNumberOfComponents()` — the same expression already used to construct `SysNewtonRhapsonPhaseEnvelope` — so the allocation and the loop bound come from one source and cannot drift apart again.

Allocation happens in `traceEnvelope()` and on the second-pass reset, the two places the arrays were previously sized.

The fixed 100 was also inconsistent with the rest of the engine, which documents its own cap:

```java
// ThermodynamicModelSettings
/** Constant <code>MAX_NUMBER_OF_COMPONENTS=200</code>. */
int MAX_NUMBER_OF_COMPONENTS = 200;
```

So the envelope was capping fluids at half the limit the thermodynamic layer supports.

## Reproduction

A light gas padded with TBP fractions, over the boundary:

| components | before | after |
|---|---|---|
| 99 | completed | completed |
| 100 | completed | completed |
| 101 | **ArrayIndexOutOfBoundsException** | completed |
| 105 | **ArrayIndexOutOfBoundsException** | completed |

**Behaviour is unchanged below the boundary.** At 100 components the envelope returns cricondentherm 166.3 °C at 82.0 bara both before and after.

## Verification

```
PTPhaseEnvelopeComponentLimitTest     3 tests   (new)
PTPhaseEnvelopeMichelsenTest         20 tests
PTPhaseEnvelopeRobustnessTest        38 tests
PTPhaseEnvelopeTest                   9 tests
RobustPhaseEnvelopeTest               5 tests
PTPhaseEnvelopeReferenceEosTest       2 tests
PTPhaseEnvelopeSegmentsTest           2 tests
PhaseEnvelopeZeroCheckTest            2 tests
------------------------------------------------
81 tests, 0 failures, BUILD SUCCESS
```

The new test covers three things: that a 120-component fluid does not throw, that the four criconden composition arrays have exactly one entry per component rather than a fixed 100, and that a 100-component fluid still works so the boundary case is not disturbed.

`spotless:apply` run.

## Two things found alongside, not addressed here

**A second wall at 200 components.** Beyond that, `addComponent` itself fails:

```
neqsim.util.exception.InvalidInputException: PhaseSrkEos:addComponent
  - Input compNumber must be valid index, i.e., between 0 and 200.
```

That is the documented `MAX_NUMBER_OF_COMPONENTS`, applied consistently across `Component`, `ComponentEos` and `UNIFACgroup`, so raising it is a much larger change than this one and is left alone.

**Empty envelopes at high component counts.** At 150 components the trace completes but returns zero points, and `getCricondenTherm()` reports the fluid's own inlet temperature and pressure rather than a computed value. That is the same silent-failure signature as #3739 and is independent of this fix — it reproduces on a plain three-component methane/ethane/propane gas.


EvenSol:
Resolved the merge conflict with `master` in cce2b271057612d2b096348d10d9c61ddee2a9a9.

The resolution preserves the convergence and stale-result protections from #3739 while keeping this PR's component-sized composition arrays. `clearResults()` and the second tracing pass now use the same allocation helper. The existing component-limit tests are unchanged.

Validation on the resolved tree (Java 17):
- `./mvnw -B -ntp -Djacoco.skip=true -Dtest=PTPhaseEnvelopeComponentLimitTest,PTPhaseEnvelopeMichelsenTest,PTPhaseEnvelopeRobustnessTest,UMRPRUPhaseEnvelopeRegressionTest test` — exit 0; 78 tests, 0 failures, 0 errors, 1 skipped.
- `python devtools/run_spotless.py apply` — exit 0, repeated with no additional changes.
- `python devtools/run_spotless.py check` — exit 0.
- `python devtools/check_documentation_search.py` — exit 0.
- `python -m pre_commit run --all-files --hook-stage pre-commit` — exit 0.
- `python -m pre_commit run --all-files --hook-stage pre-push` — exit 0.
- `git diff --cached --check` — exit 0.

Python commands used the Work runtime interpreter; Maven was bootstrapped with the NeqSim skill helper. The published tree matches the validated local tree exactly (`96609ee1bc81fcadb46e62071b7c4d372ff57260`).

Documentation impact: none beyond correcting the allocation comment; the current phase-envelope convergence documentation remains valid. GitHub CI is pending for the new merge commit.

PR #3826, opened 2026-09-20
Fix SessionStart hook: encode Windows PowerShell command to survive $-token stripping

EvenSol:
## Problem

The generated plugin `SessionStart` hook's `windows` command in `hooks.json` (produced by `devtools/build_agent_plugin.py`) is mangled before PowerShell ever sees it. Whatever resolves the hook's `windows` string before invoking it strips **any** literal `$identifier` / `${identifier}` token — not just `${PLUGIN_ROOT}` — replacing each with an empty string. That turned the self-locating inline script:

```
$r = '${PLUGIN_ROOT}'; if (-not $r -or $r.Contains('{PLUGIN_ROOT}')) { $r = $env:PLUGIN_ROOT }; ...
```

into garbage like:

```
 = ''; if (-not  -or .Contains('{PLUGIN_ROOT}')) {  =  }; ...
```

producing three PowerShell parse errors per VS Code session (one per installed plugin: `neqsim`, `neqsim-community`, `neqsim-enterprise`), and silently skipping the skill-package install every time.

## Fix

`hook_commands()` now ships the real PowerShell script as a base64 (UTF-16LE) `-EncodedCommand` payload instead of an inline `-Command "..."` string. Base64 has no `$`, so whatever strips `$`-tokens has nothing to touch, and PowerShell decodes/executes the intact script. The now-unnecessary literal `${PLUGIN_ROOT}`-placeholder check is dropped (a spec-conformant client also sets the env var, which the script still checks first). The POSIX `sh` command is unchanged — not exercised on Windows and no evidence of the same issue there.

`test_build_agent_plugin.py` is updated to assert `"$" not in windows` and to decode the `-EncodedCommand` payload to check the actual script content.

## Verification

- `pytest devtools/test_build_agent_plugin.py` — 14 passed.
- Regenerated the deployed `hooks.json` for all 3 plugins (both the `neqsim-copilot-plugin` checkout and the installed `~/.vscode/agent-plugins/...` copy) and ran the resulting `windows` command via `cmd /c $winCmd` both with `PLUGIN_ROOT` preset and unset (self-locate via `Get-ChildItem`) — exit code 0 in all cases, no parse errors, self-locate correctly finds the installed plugin folder and starts the install.


EvenSol:
Fixed the failing agent/skill cross-reference check in 2632f52. Both referenced skills exist in `equinor/neqsim-community-skills` (`skills/process/neqsim-cfd-coupling` and `skills/subsurface/neqsim-near-well-and-injectivity`); their ownership is now declared in `.github/external-skill-refs.txt`.

Validation passed: reference verifier (four existing warnings), 14 build-agent-plugin tests, direct Spotless apply/check, documentation search audit, and both pre-commit/pre-push hook stages. Documentation impact: none; the existing links are correct. Fresh GitHub CI is pending.

Issue #3822, opened 2026-09-20
NeqSim defects from a software HAZOP of the calculation layer


PatrickMockridge:
**Provenance.** These come from a software HAZOP of NeqSim's calculation layer (7 guidewords applied to each safeguard
as a node), not from the port. Where a finding sits underneath an issue that is already filed, it is marked
**[RELATED]** and carries this caveat:

> **This issue comes from a HAZOP run to elucidate root causes. It is *extra* to the previously raised issue(s) it
> references. It does not supersede them, does not dispute their fixes, and does not ask for them to be reopened.
> Where the earlier issue is closed, it stays closed; this is a distinct defect or a root cause beneath it.**

**Environment.** NeqSim at HEAD, `pom.xml` version **3.21.0**, `src/main/resources/data/COMP.csv` at **389 rows /
171 columns**. The version matters: several of the eleven issues filed against `neqsim-3.20.0` no longer reproduce
here, and two of the findings below exist *because* of a fix that landed since.

**Verification status.** Every finding states how it was established. No JVM was available in the environment, so
nothing here is a run. The categories are:

| Mark | Meaning |
|---|---|
| **ARITHMETIC** | The value follows from the shipped coefficients and the expression quoted from source. Computed exactly; anyone can reproduce it with the row and the line. |
| **SOURCE** | Provable by reading the quoted lines. No run needed — it is a code defect, not a numerical one. |
| **COUNT** | A count over `COMP.csv`, reproducible with the script given. |

One candidate finding was **withdrawn** during verification and is recorded here rather than filed: an apparent bypass
of the ion guard by `H+PZCOO-` (`COMPTYPE=ion`, `IONICCHARGE=0`). It does not bypass. The guard is
`ionicCharge == 0 && !isIsIon()`, and `isIsIon()` reads `componentType`, which is `ion`
(`Component.java:2126-2130`), so the row is correctly excluded. The initial reading checked one condition of a
two-condition guard. It is recorded because a reader may draw the same conclusion and should not.

---

## Issue 1 — `getAntoineVaporPressure` returns `+Infinity` for `H2O2` **[RELATED — #3768]**

> **HAZOP caveat.** This comes from a HAZOP run to elucidate root causes. #3768 identified the unreachable DIPPR-101
> branch and its fix is correct — `usesDipprVaporPressureCorrelation()` now selects DIPPR for rows labelled `log`
[...]

PR #3830, opened 2026-09-20
[AI-Assisted] fix(thermo): include complete Mie hard-chain thermal dependence

### LINKED ISSUE #3814: `PhaseSAFTVRMie.dF_HC_SAFTdT()` omits the chain contact value's temperature dependence, so `Hres`/`Sres` are wrong for every fluid with `m > 1`

`neqsim/thermo/phase/PhaseSAFTVRMie.java`, `dF_HC_SAFTdT()` (~line 2182).

---

## Summary

`dF_HC_SAFTdT()` differentiates the chain contact value `g_hs` as though it were a function
of the packing fraction alone. It is not, so the method is wrong for every fluid whose
components have more than one segment — that is, every real fluid. On n-butane alone at
350 K and 30 bara its answer is not merely the wrong size: **it has the wrong sign.**

`Hres`/`Sres` at that state are 6.7% low. `dFdT()` is 8.0% low.

**The volume derivatives are not affected.** At constant temperature the contact value
really does move with the packing fraction alone, so `dF_HC_SAFTdV()`, the volume solve, the
pressure and the fugacity coefficients are all right.

---

## Why

```java
public double dF_HC_SAFTdT() {
  return getNumberOfMolesInPhase() * (mSAFT * daHSSAFTdN * dNSAFTdT - mmin1SAFT / ghsSAFT * dgHSSAFTdN * dNSAFTdT);
}
```

`dNSAFTdT` is `d eta/dT` at constant volume, so both terms are the chain term's derivative
**along the packing fraction**. That is the whole derivative only if `g_hs` depends on the
packing fraction and nothing else — and `volInit()` replaces the Carnahan-Starling contact
value with

```
ghsSAFT = exp( sum_i w_i ln g_Mie_ii / W ),   w_i = x_i (m_i - 1)
g_Mie_ii = g_HS0_i exp( beta_i (g_1 + beta_i g_2) / g_HS0_i ),   beta_i = eps_i / (k T)
```

`calcLnGChainEffective(etaVal, alpha)` reads `ci.getEpsikSAFT() / temperature` beside
`etaVal` and `ci.getdSAFTi()` beside it, so `g_hs` is a function of three things and the
class carries one of them:

| route | carried by `dF_HC_SAFTdT()`? |
|---|---|
| `T` through `eta`, via `dgHSSAFTdN * dNSAFTdT` | yes |
| `T` at fixed `eta`, through `beta = eps/(kT)` | **no** |
| `T` through each segment diameter `d_i(T)` | **no** |

The two missing routes are not a correction — they are most of the derivative. On
methane/n-butane at 350 K and 30 bara, `g_hs` itself moves by `-1.71e-04` per K, where
`dgHSSAFTdN * dNSAFTdT` accounts for `-6.93e-06` of it. The rest is the two routes above.

`dgHSSAFTdN` itself is not at fault: evaluating `calcLnGChainEffective` by hand at `eta` and
`eta ± |eta| 1e-4` reproduces it to twelve digits. The difference quotient is faithful to a
function that is not the one `F_HC_SAFT()` evaluates.

---

## How to reproduce

Self-contained; no test data needed. The trick is that `F_hc` is differenced at a **pinned
molar volume** — perturb `T`, restore the volume, re-run `volInit()` — so the comparison
reaches none of the derivative code above. `setMolarVolume(v)` + `volInit()` is the pair the
class's own `dFdVdVdV()` already uses for exactly this purpose.

```java
import neqsim.thermo.phase.PhaseSAFTVRMie;
import neqsim.thermo.system.SystemSAFTVRMie;

public class MieChainD {
  private static PhaseSAFTVRMie pinned(double t, double v) {
    SystemSAFTVRMie s = new SystemSAFTVRMie(t, 30.0);
    s.addComponent("n-butane", 1.0);
    s.setMixingRule("classic");
    s.init(0);
    s.init(1);
    s.init(2);
    PhaseSAFTVRMie p = (PhaseSAFTVRMie) s.getPhase(0);
    p.setMolarVolume(v);
    p.volInit();
    return p;
  }

    SystemSAFTVRMie s = new SystemSAFTVRMie(350.0, 30.0);
    s.addComponent("n-butane", 1.0);
    s.setMixingRule("classic");
    s.init(0);
    s.init(1);
    s.init(2);
    PhaseSAFTVRMie base = (PhaseSAFTVRMie) s.getPhase(0);
    double v = base.getMolarVolume();
    double h = 0.05;

    System.out.printf("reported dF_HC_SAFTdT() = %.15g%n", base.dF_HC_SAFTdT());
    System.out.printf("its own F_hc, fixed V   = %.15g%n",
        (pinned(350.0 + h, v).F_HC_SAFT() - pinned(350.0 - h, v).F_HC_SAFT()) / (2.0 * h));
    System.out.printf("reported dFdT()         = %.15g%n", base.dFdT());
    System.out.printf("its own F, fixed V      = %.15g%n",
        (pinned(350.0 + h, v).getF() - pinned(350.0 - h, v).getF()) / (2.0 * h));
    System.out.printf("reported HresTP()/n     = %.15g%n",
        base.getHresTP() / base.getNumberOfMolesInPhase());
    System.out.printf("m (segments)            = %.15g%n", base.getmSAFT());
  }
}
```

Output as it stands:

```
reported dF_HC_SAFTdT() = -6.58222781509927e-05
its own F_hc, fixed V   = 0.000180909655021488
reported dFdT()         = 0.00284373111666029
its own F, fixed V      = 0.00309046312607308
reported HresTP()/n     = -3522.40481784965
m (segments)            = 1.85140000000000
```

## Expected

`+1.80909655021488e-04` for the chain term, and `0.00309046312607308` for `dFdT()`.

The difference is not step-sensitive. For `F_hc` it is `1.80905497454420e-04`,
`1.80909655021488e-04` and `1.80910022025832e-04` at `h` of `0.005`, `0.05` and `0.2` K —
converged to eight digits, and the reported value is not near any of them.

## Actual

`-6.58222781509927e-05` for the chain term, and `0.00284373111666029` for `dFdT()`.

The reported chain term is `-0.36` times the correct one — **the sign is wrong** — and
`dFdT()` is 8.0% low. Through `Hres/(RT) = Z − 1 − T·dFdT` that makes `getHresTP()/n` `-3522.4 J/mol` where
the class's own energy gives `-3773.7 J/mol`, a 6.7% error.

## Methane is the exception, and it is easy to test on

At `m = 1` the block

```java
if (mmin1SAFT > 1.0e-10 && eta > 1.0e-10 && eta < 0.55) { ... }
```

is false, so `ghsSAFT` keeps the Carnahan-Starling value — which *is* a function of the
packing fraction alone — and `dF_HC_SAFTdT()` then agrees with its own `F` to `4e-8`
relative. Pure methane is the easy state to reach for, and it is the one state where the
bug is invisible.

Measured on the same three states, each at `h = 0.01` K:

| state | reported | difference of its own `F_hc` |
|---|---|---|
| methane, `m = 1` | -3.69985758564022e-05 | -3.69985760007019e-05 |
| n-butane, `m = 1.8514` | -6.58222781509927e-05 | **+1.80907699712241e-04** |
| methane/n-butane, 0.6/0.4 | -3.09846277671377e-05 | **+2.34611958374598e-05** |

---

## Fix

`dF_HC_SAFTdT()` needs the full `d g_hs/dT`, not only its packing-fraction route. The
smallest patch keeps the class's existing shape: in `volInit()`, beside `dgHSSAFTdN`, store
a second central difference of `calcLnGChainEffective` in `T` at a fixed packing fraction
with the diameters held — call it `dLogGhsDTSAFT` — and then

```java
double dLogGhsDT = dgHSSAFTdN / ghsSAFT * dNSAFTdT + dLogGhsDTSAFT;
return getNumberOfMolesInPhase() * (mSAFT * daHSSAFTdN * dNSAFTdT - mmin1SAFT * dLogGhsDT);
```

The third route — each diameter's own `T d ln d_i/dT` — is not in that difference and has to
be carried separately. It is the one that makes the sign come out right on n-butane.

`dF_HC_SAFTdTdV()` and `dF_HC_SAFTdTdT()` carry the same `dNSAFTdT` term and are wrong the
same way.

EvenSol:
The Mie hard-chain thermal derivatives treated the contact value as depending only on packing fraction and returned zero diameter curvature. This omits the explicit epsilon/(kT) dependence and part of the Barker–Henderson diameter dependence.

Evaluate the complete hard-chain energy with consistent trial temperature and diameters for dT, dTT and dTV, using finite-difference stencils that do not mutate the phase. Evaluate the second diameter-moment derivative from the existing BH quadrature. Existing hard-chain blend/domain behavior is retained.

Validation: the regression fails on unmodified master. This PR alone passes 58 tests, including the existing SystemSAFTVRMie suite. Independent energy perturbations cover methane, n-butane and their mixture at 300 and 350 K, total dT, curvature, the mixed derivative and preservation of the receiver state. The mixed check uses a separate, larger four-corner energy stencil to avoid differentiating the already numerical volume slope.

Documentation: method Javadocs describe the fixed-volume basis and numerical evaluation. These are numerical derivatives, with finite-difference cost and approximation error; they are not claimed as closed-form expressions or validation outside the tested states. No new user workflow or guide is required.

Closes #3814
Addresses #3822 finding 7 (Mie diameter curvature).

Reproduce the focused check: `./mvnw test -Dtest='SystemSAFTVRMieTest,MieDerivativeConsistencyTest'`. Java 8 target, JDK 17, NeqSim 3.21.0. The combined selected Maven lifecycle passed 343 tests (two pre-existing skips). Spotless apply was repeated to stability, Spotless check and the documentation-search audit passed. The full repository test suite was not run locally. No changes to default branch or automatic merging are requested.

## CI repair — reservoir scheduling (2026-09-20)

Commit `11a39d1f7102dd78d1565f139bb14723ba88176c` fixes the producer flash race observed in `ProcessModelOperatingActionEvaluatorTest`. `SimpleReservoir` now reports its gas/oil/water producers and gas/water injectors through the standard stream topology API. The parallel scheduler therefore places separately registered producer streams after their reservoir instead of flashing the same mutable fluid concurrently.

Added a regression that failed before the fix and verifies all five stream connections, defensive list containers, dependency levels and parallel execution. `./mvnw -B -ntp -Dtest=SimpleReservoirStreamTopologyTest,ProcessModelOperatingActionEvaluatorTest,ProcessModelOperatingActionSetEvaluatorTest,SimpleReservoirTest,MieDerivativeConsistencyTest test` passed: 12 tests, zero failures/errors on Java 17. Documentation impact: updated the reservoir guide and Javadocs to explain stream ownership and scheduling.

After Maven bootstrap, `python devtools/run_spotless.py apply` (clean repeat), `python devtools/run_spotless.py check`, `python devtools/check_documentation_search.py`, both `pre-commit run --all-files --hook-stage pre-commit` and `--hook-stage pre-push`, and `git diff --cached --check` all exited 0. **VALIDATION PENDING CI** on the new head.


EvenSol:
CI follow-up: the original slow partition 3 failed in `BottleneckAnalysisOptimizerTest.testTwoStageOptimizationRecommendedApproach` (compressor utilization 100.7%/101.6%). That fixture creates `SystemPrEos`; the original numerical change was limited to `PhaseSAFTVRMie`.

Evidence correction (2026-09-20): my earlier statement that the original-master optimizer class passed locally was not supported by that command. It used `-DexcludedGroups=` instead of this repository's `-DexcludedTestGroups=`, so the slow-tagged class was excluded. That run must not be cited as a passing baseline test.

The subsequent repair in ce62b152c383cbac7df0da3bf8e44b07b3f5e0bc brackets a feasible compressor operating point and retains/tightens the physical assertions. Its current Java 21 CI evidence is verified: job 106078467394 ran all 10 optimizer tests successfully; the complete slow partition ran 103 tests with 0 failures, 0 errors and 2 existing skips. The original failed head is superseded and does not need a rerun. Other current-head CI jobs remain in progress.

EvenSol:
Fixed the failing optimization test in ce62b15. The fixture used binary feasibility search over a compressor range containing both low-flow surge/minimum-speed limits and high-flow capacity limits. It now probes single operating points to locate a verified feasible bracket, then bisects the upper boundary. The capacity assertion is tightened from 102% to 100%, with no changes to the Mie implementation or production optimizer.

Updated the optimization guidance and applied the release documentation correction from #3836 because this branch now triggers the documentation job. All 10 BottleneckAnalysisOptimizerTest methods and MieDerivativeConsistencyTest pass locally; 385 documentation tests, direct Spotless apply/check, search audit, and both hook stages pass. Local JDK 17; fresh Java 21 CI is pending.

PR #3833, opened 2026-09-20
[AI-Assisted] fix(flash): restore saturation failure propagation

EvenSol:
Eleven saturation wrappers declared IsNaNException but left their failure throws commented out. The boolean bubble-pressure wrapper also caught solver exceptions and returned normally.

Restore the guards, reject nonfinite/nonpositive solved values and reported supercritical outcomes, and propagate the caught bubble-pressure failure with its original cause. Check temperature, rather than pressure, after a constant-phase-fraction temperature solve.

Validation: the public boolean bubble wrapper regression fails on unmodified master. This PR alone passes 25 tests covering failure propagation, valid bubble/dew states, supercritical rejection, pure-component saturation and capillary dew points.

Documentation impact: existing method signatures/Javadocs already advertise failure exceptions. This restores that contract; callers that relied on silent failed solves will now receive an exception. Solver convergence thresholds are unchanged.

Addresses #3822 finding 2.

Reproduce the focused check: `./mvnw test -Dtest='SaturationFailurePropagationTest,SuperCriticalSaturationOpsTest,PureComponentSaturation*Test,CapillaryDewPointFlashTest'`. Java 8 target, JDK 17, NeqSim 3.21.0. The combined selected Maven lifecycle passed 343 tests (two pre-existing skips). Spotless apply was repeated to stability, Spotless check and the documentation-search audit passed. The full repository test suite was not run locally. No changes to default branch or automatic merging are requested.


EvenSol:
Fixed the CI failures exposed by restored saturation error propagation in 955fe93337f24c2c405ac2746fccb5c90b4eb43d.

Root causes and changes:
- Bubble-temperature, bubble-pressure, and dew-temperature iterations could collapse to trivial equal-composition roots from otherwise valid starting conditions. Added bounded Wilson-based restart estimates; the EOS still determines the equilibrium, and unsuccessful calculations still propagate errors.
- UMR-CPA treated CO2 association-site metadata as proof of fitted CPA physical parameters, assigning zero energy/co-volume values. It now retains PR physical parameters unless both CPA parameters are finite and positive.
- The wet-oil documentation test added water after setting the mixing rule, leaving stale binary interaction matrices. Rebuilt the rule after the composition change and documented this requirement.
- Updated the five stale current-version documentation examples to 3.21.0 for the documentation CI gate.

Validation: reproduced the original condenser and nine regular-suite failures locally before fixing them. The combined targeted suite now reports 174 tests, 0 failures, 0 errors, and 2 existing skips. This covers the failing callers, distillation, saturation/capillary/supercritical cases, and UMR-CPA phase, mixture, volume, and dehydration tests. New regressions check phase separation and fugacity equality across poor initial guesses; pure UMR-CPA benchmarks also verify actual equilibrium. All 385 documentation tests, source audit, Spotless apply/check, and both pre-commit stages passed. Full GitHub CI is running on the new head.

PR #3838, opened 2026-09-20
[AI-Assisted] Fix dilute Pitzer attenuation and document neutral support

### LINKED ISSUE #3825: 28 of the bundled Pitzer catalogue's 37 neutral-interaction rows name a species that is not a component, so the λ/ζ/μ/η layer activates for one chemistry only

`neqsim/thermo/phase/PitzerParameterDatasets.java`, `validateCatalogTopology` and
`applyCatalogNeutralRows`.
`src/main/resources/neqsim/thermo/phase/phreeqc/pitzer-b0b3be76….dat`, the `-LAMBDA` and
`-ZETA` sections.

---

## Summary

`tryApplyCompletePhreeqcPitzerCatalog` applies the catalogue only when it covers the
phase's **whole** topology, and for a neutral solute that means a `LAMBDA` row against
every other neutral and every ion, plus a `ZETA` row against every cation-anion pair. The
lookups are keyed on the **component's name**:

```java
phase.setLambdaTemperatureCoefficients(neutral, ion, PHREEQC_REFERENCE_TEMPERATURE_K,
    catalog.require(Family.LAMBDA, neutralName, phase.getComponent(ion).getComponentName()));
```

The catalogue's neutral sections name five species that NeqSim's component database does
not carry under any spelling:

| the catalogue writes | a component? |
|---|---|
| `B(OH)3` | no — no boron species at all |
| `H4SiO4` | no — no silica species at all |
| `H2Sg` | no — the database's row is `H2S` |
| `(H2Sg)2` | no |
| `Hdg` | no — the database's row is `hydrogen` |

**So of the 37 neutral-family rows, 28 can never be looked up**, and the nine that can are
all `CO2`:

```
LAMBDA  CO2|CO2   CO2|Na+   CO2|K+   CO2|Ca++   CO2|Mg++   CO2|Cl-   CO2|SO4--   CO2|HSO4-
ZETA    CO2|Na+|SO4--
```

**The neutral layer therefore activates for exactly one chemistry: CO₂ with sodium and
sulphate.** Measured:

| mixture | dataset applied | neutral layer |
|---|---|---|
| `water + Na+ + SO4-- + CO2` | PHREEQC catalogue | **active** — λ contributions to Na⁺, SO₄²⁻ and CO₂ |
| `water + Na+ + Cl- + CO2` | legacy fallback | inactive |
| `water + Na+ + Cl- + H2S` | legacy fallback | inactive |

The chloride case is a *different* matter and is documented — `applyPhreeqcCo2ChlorideParameters`
says the bundled source has no CO₂–cation–Cl ζ rows and requires the caller to supply them.
The H₂S case is not documented anywhere and looks structural: no spelling NeqSim's database
uses reaches `H2Sg`, so the four sulphide `ZETA` rows and their `LAMBDA` siblings are dead.

---

## How to reproduce

Self-contained; no test data needed. It performs the lookups a phase performs, and then
builds the phase to show which dataset it gets.

```java
import java.util.Map;
import neqsim.thermo.phase.PhasePitzer;
import neqsim.thermo.phase.PhreeqcPitzerParameterCatalog;
import neqsim.thermo.system.SystemPitzer;

public class NeutralRows {
  private static void lookup(PhreeqcPitzerParameterCatalog c, String family, String... species) {
    double[] found = c.find(PhreeqcPitzerParameterCatalog.Family.valueOf(family), species);
    System.out.printf("  %-7s %-24s %s%n", family, String.join(", ", species),
        found == null ? "MISSING" : "present");
  }

  private static void phase(String label, String[] names, double[] z) {
    SystemInterface s = new SystemPitzer(298.15, 1.0);
    for (int i = 0; i < names.length; i++) {
      s.addComponent(names[i], z[i]);
    }
    s.setMixingRule("classic");
    s.init(0);
    s.init(1);
    PhasePitzer p = (PhasePitzer) s.getPhase(1);
    System.out.printf("  %-30s dataset=%-7s neutral layer active=%s%n", label,
        p.getParameterDatasetId().startsWith("usgs") ? "phreeqc" : "legacy",
        p.hasNeutralPitzerInteractions());
  }

    PhreeqcPitzerParameterCatalog c = PhreeqcPitzerParameterCatalog.getInstance();

    System.out.println("the rows the catalogue carries for those species:");
    lookup(c, "LAMBDA", "H2Sg", "Na+");
    lookup(c, "LAMBDA", "H2Sg", "H2Sg");
    lookup(c, "ZETA", "H2Sg", "Na+", "Cl-");
    lookup(c, "LAMBDA", "B(OH)3", "Na+");
    lookup(c, "LAMBDA", "H4SiO4", "Na+");
    lookup(c, "LAMBDA", "CO2", "Na+");
    lookup(c, "ZETA", "CO2", "Na+", "SO4--");

    System.out.println("\nand the phases those rows are meant for:");
    phase("water + Na+ + Cl- + H2S", new String[] {"water", "Na+", "Cl-", "H2S"},
        new double[] {0.86, 0.03, 0.03, 0.08});
    phase("water + Na+ + SO4-- + CO2", new String[] {"water", "Na+", "SO4--", "CO2"},
        new double[] {0.86, 0.06, 0.03, 0.05});
  }
}
```

Output as it stands:

```
the rows the catalogue carries for those species:
  LAMBDA  H2Sg, Na+               present
  LAMBDA  H2Sg, H2Sg              present
  ZETA    H2Sg, Na+, Cl-          MISSING
  LAMBDA  B(OH)3, Na+             present
  LAMBDA  H4SiO4, Na+             present
  LAMBDA  CO2, Na+                present
  ZETA    CO2, Na+, SO4--         present

and the phases those rows are meant for:
  water + Na+ + Cl- + H2S        dataset=legacy  neutral layer active=false
  water + Na+ + SO4-- + CO2      dataset=phreeqc neutral layer active=true
```

The three `present` lines are the point: the catalogue *has* the data, and the phase still
falls back, because `H2Sg` is not a name `getComponentName()` can return.

---

## Expected

Either the rows are reachable — in which case `water + Na+ + Cl- + H2S` should take the
catalogue and report an active neutral layer — or they are not, and the situation is
documented the way the CO₂–chloride gap is.

## Actual

The rows are present in the catalogue and unreachable from any phase.

---

## What would settle it

The species fall into two groups and they need different answers.

**`H2Sg`, `(H2Sg)2` and `Hdg` have plausible NeqSim counterparts** — `H2S` and `hydrogen`
are both components. Is a name mapping intended (`H2Sg` ⇒ `H2S`), or are the PHREEQC
aqueous-gas species names deliberately distinct from the components because they mean
different things? `ComponentGE.fugcoef` and `PhasePitzer` would both need the answer.

**`B(OH)3` and `H4SiO4` have no counterpart at all** — the database carries no boron and no
silica species. Rows for boron and silica chemistry cannot be reached without adding
components, which is a much larger question than this file.

If the intent is that only CO₂ is supported and the rest are carried for completeness, then
the file is honest as it stands and this is a note rather than a defect — but the
`-LAMBDA` and `-ZETA` sections would read as available data to anyone looking at them, and
`hasNeutralPitzerInteractions()` returning `false` is the only sign otherwise.

### LINKED ISSUE #3827: `ComponentGePitzer` guards Pitzer's `g(x)` at `1e-12` and returns cancellation noise, while `PitzerBinaryVolumetricModel` implements the same function correctly at `1e-4` and a test asserts it

`neqsim/thermo/component/ComponentGePitzer.java` lines 257, 263, 271, 276, 395, 401.
`neqsim/thermo/phase/PitzerBinaryVolumetricModel.java` line 170.
`src/test/java/neqsim/thermo/phase/PitzerBinaryVolumetricModelTest.java` line 39.

---

## Summary

Pitzer's attenuation function

```
g(x) = 2 (1 - (1 + x) exp(-x)) / x^2,   g(0) = 1
```

is implemented **twice in this repository, with different answers.**

`PitzerBinaryVolumetricModel` gets it right. It switches to the Maclaurin series below
`1e-4`, and its test asserts the small-argument limit to `1e-16`:

```java
@Test
void evaluatesAttenuationWithoutDiluteCancellation() {
  assertEquals(1.0, PitzerBinaryVolumetricModel.attenuation(0.0), 0.0);
  assertEquals(1.0 - 2.0e-12 / 3.0, PitzerBinaryVolumetricModel.attenuation(1.0e-12), 1.0e-16);
```

`ComponentGePitzer` guards the *direct* expression at `x > 1e-12` and returns `0.0` below
that. The direct expression is a cancellation of two numbers near one — the numerator is
`x^2/2 - x^3/3 + x^4/8 - ...`, so it loses about sixteen significant digits — and by
`1e-12` it has nothing left. **The two classes return different values at the same
argument, and the test is the one that is right.**

The phase model's copy appears at **six sites**, all with the same threshold:

| line | guard | expression |
|---|---|---|
| 257 | `x1 > 1e-12` | `g1`, `gp1` (258, 259) |
| 263 | `I > 1e-12` | `Bprime = beta1 * gp1 / I` |
| 271 | `x2 > 1e-12` | `g2`, `gp2` (272, 273) |
| 276 | `I > 1e-12` | `Bprime += beta2 * gp2 / I` |
| 395 | `x1 > 1.0e-12 && ionicStrength > 1.0e-12` | `gp1` in `phreeqcBinaryBprime` (396) |
| 401 | same, plus the `beta2` test | `gp2` (402) |

Fixing the two in `getGamma` and not the two in `phreeqcBinaryBprime` leaves the catalogue
path wrong, so the list matters.

---

## Measured

Against the series `1 - (2/3)x + x^2/4 - x^3/15`, which agrees with the direct expression
to `1e-10` at `1e-3` and `1e-4` — the series' own truncation, not either method's error:

| x | `g(x)` from the direct expression | the true value | relative error |
|---|---|---|---|
| 1e-3 | 0.999333583440 | 0.999333583267 | 1.7e-10 |
| 1e-4 | 0.999933336132 | 0.999933335833 | 3.0e-10 |
| **1e-5** | 0.999991200956 | 0.999993333358 | **2.1e-6** |
| **1e-6** | 1.000088900582 | 0.999999333334 | **9.0e-5** |
| **1e-7** | 0.976996261670 | 0.999999933333 | **2.3e-2** |
| **1e-8** | 2.220446049250 | 0.999999993333 | **1.22** |
| **1e-9** | 0.000000000000 | 0.999999999333 | **1.00** |

From `1e-5` the value is imprecise; below `1e-8` it is arbitrary — `2.220446` at `1e-8` is
a different number, and from `1e-9` the numerator rounds to zero and the returned `0.0` is
indistinguishable from the guard's.

**`gp1` is worse than `g1`, and it is the one that is divided by `I`.** The bracket in
`1 - (1 + x + x^2/2) exp(-x)` cancels from its *cubic* term, not its quadratic one:

| x | `g'(x)` from the direct expression | the true value | relative error |
|---|---|---|---|
| 1e-4 | -3.332889519925e-05 | -3.333083343333e-05 | 5.8e-5 |
| **1e-5** | -2.220446049250e-06 | -3.333308333433e-06 | **33%** |
| **1e-6** | 0.0 | -3.333330833334e-07 | **100%** |
| **1e-7** | **+4.440892098501e-02** | -3.333333083333e-08 | **-1.3e6, wrong sign** |

At `1e-7` the computed derivative is thirteen thousand times the true magnitude and
positive where the truth is negative, and it is then divided by `I`.

---

## How to reproduce

Self-contained; no test data needed.

```java
public class GGuard {
  static double neqsim(double x) {
    if (x > 1e-12) {
      return 2.0 * (1.0 - (1.0 + x) * Math.exp(-x)) / (x * x);
    }
    return 0.0;
  }

  static double neqsimGp(double x) {
    if (x > 1e-12) {
      return -2.0 * (1.0 - (1.0 + x + x * x / 2.0) * Math.exp(-x)) / (x * x);
    }
    return 0.0;
  }

  // The same functions by their Maclaurin series, which have no cancellation:
  //   g  = 1 - (2/3)x + x^2/4 - x^3/15
  //   g' = -x/3 + x^2/4 - x^3/10
  static double series(double x) {
    return 1.0 - (2.0 / 3.0) * x + 0.25 * x * x - (1.0 / 15.0) * x * x * x;
  }

  static double seriesGp(double x) {
    return -x / 3.0 + 0.25 * x * x - x * x * x / 10.0;
  }

    System.out.printf("%8s %22s %22s %11s | %22s %22s %11s%n",
        "x", "g", "series", "rel err", "g'", "series", "rel err");
    for (double x : new double[] {1e-3, 1e-4, 1e-5, 1e-6, 1e-7, 1e-8, 1e-9}) {
      System.out.printf("%8.0e %22.12f %22.12f %11.2e | %22.12e %22.12e %11.2e%n", x,
          neqsim(x), series(x), Math.abs(neqsim(x) - series(x)) / series(x),
          neqsimGp(x), seriesGp(x), Math.abs(neqsimGp(x) - seriesGp(x)) / Math.abs(seriesGp(x)));
    }
  }
}
```

The second half of the last row is the clearest statement of the bug: at `1e-12`, where the
test in this repository asserts `1 - 2e-12/3`, `ComponentGePitzer`'s guard returns `0.0`.

---

## Expected

One function, one answer. `g` rises to `1` as `x` falls to `0` from below, in both classes,
to whatever precision the series gives.

## Actual

`PitzerBinaryVolumetricModel.attenuation` gives it; `ComponentGePitzer` gives `0.0`,
`2.220446`, and a sign-flipped `g'`.

---

## Fix

`PitzerBinaryVolumetricModel.attenuation` is the implementation to copy — same threshold,
same series, already written and already tested:

```java
if (x < 1.0e-4) {
  return 1.0 + x * (-2.0 / 3.0 + x * (1.0 / 4.0 + x * (-1.0 / 15.0 + x / 72.0)));
}
```

At the six sites in `ComponentGePitzer`, that is

```java
double g1;
double gp1;
if (x1 < 1.0e-4) {
  g1 = 1.0 + x1 * (-2.0 / 3.0 + x1 * (1.0 / 4.0 - x1 / 15.0));
  gp1 = x1 * (-1.0 / 3.0 + x1 * (1.0 / 4.0 - x1 / 10.0));
} else {
  g1 = 2.0 * (1.0 - (1.0 + x1) * Math.exp(-x1)) / (x1 * x1);
  gp1 = -2.0 * (1.0 - (1.0 + x1 + x1 * x1 / 2.0) * Math.exp(-x1)) / (x1 * x1);
}
```

Three terms suffice at `1e-4`: the `x^4` term is under `1e-18`. The two `I > 1e-12` guards
on `Bprime` can then go, or stay as they are; they are unreachable from either direction
once `gp1` is sound.

---

## Reachability, and why it does not decide this

`x = alpha * sqrt(I)` with `alpha` in `{1.4, 2.0}`, so `x < 1e-5` needs an ionic strength
below `2.5e-11 mol/kg` — a sodium chloride mole fraction near `1e-13`. The terms affected
at that dilution are already vanishingly small: the change this makes to `ln gamma` is at
the parts-in-`1e11` level, so **no ordinary brine gives a wrong answer**, and the numbers in
the table are not an impact argument.

The report is not about the size of the number. It is that the contract the guard states —
"below `1e-12` the limit is close enough to `0.0`" — is wrong by twelve orders of
magnitude, that this repository's own test contradicts it, and that the corrected form is
already sitting in the next package. A reader who fixes `ComponentGePitzer` and not
`PitzerBinaryVolumetricModel`, or the two `getGamma` sites and not the two in
`phreeqcBinaryBprime`, leaves the same function disagreeing with itself in three places.

EvenSol:
Closes #3827. Closes #3825 through its requested explicit support documentation and regression coverage.

At small positive ionic strength, direct subtraction in the Pitzer attenuation and scaled derivative loses the beta contributions; the old cutoff also gives the wrong g(0) limit. Separately, the bundled PHREEQC neutral-row inventory can be mistaken for supported NeqSim chemistry.

Changes:
- Share a stable attenuation kernel between activity and volumetric calculations, with g(0)=1 and an independently evaluated scaled derivative (x/2) dg/dx.
- Use it for beta1/beta2 in both legacy and PHREEQC common-ion activity paths. Remove the artificial positive-ionic-strength cutoff; retain the exact-zero guard in dB/dI.
- Document the 9 CO2 rows versus 28 source-only neutral rows, missing self interactions and species/speciation mappings, absence of bundled mu/eta rows, and the explicit API for rejecting unsupported coverage.
- Test legacy fallback and missing-family diagnostics without inventing aliases or fitting parameters. On current master, zeta(H2Sg,Na+,Cl-) is present and lambda(H2Sg,H2Sg) is absent, the reverse of those two lines in #3825's reported output.

Validation:
- Before the fix, the new dilute activity regression fails: expected log-gamma contribution 1.9999999966666646, actual 0.0 at molality 1e-18.
- Independent integral references cover zero, tiny arguments, the series boundary, and ordinary concentrations.
- Amplified numerical probes exercise beta1/beta2 and both common-ion modes at five concentrations; these deliberately artificial coefficients are not proposed salt fits.
- All 152 Pitzer tests pass, including existing IPhreeqc/brine and volume references. Repeated with every other proposed Patrick fix restored to the original master production code/resources: all 152 pass independently.
- Java 8 target compilation, repeated Spotless apply/check and documentation source audit pass.
- Full repository CI remains the integration gate. The shared baseline documentation-version failure is repaired separately in #3836.

AI-Assisted implementation. Based directly on master a7ad601b2478757130d7b5f2f492337effc130a4; independent of the other numerical fix PRs. Thermodynamic subsystem/domain review remains required. No changes to the archived source catalogue or unsupported-species activation policy.


EvenSol:
Fixed the failing documentation-search job in 3cde941. The five installation/quickstart pages still referenced 3.20.0 while pom.xml and the documentation contracts require 3.21.0. Applied the same correction as #3836 directly on this PR's existing head.

Validation: all 385 documentation contract tests pass; direct Spotless apply/check, documentation search audit, and both pre-commit/pre-push hook stages pass. No Java or numerical changes. Fresh GitHub CI is pending.

PR #3840, opened 2026-09-20
[AI-Assisted] Respect GE component topology and solute in infinite dilution

### LINKED ISSUE #3837: `PhaseGE.getActivityCoefficientInfDil` passes a literal `2` as the component count, so a one-component GE phase reads a component that does not exist

`neqsim/thermo/phase/PhaseGE.java` lines 197-205, `getActivityCoefficientInfDil`.

---

## Summary

The method clones the phase and evaluates it at the clone's infinite-dilution limit. It
takes the clone's own component count for the `init` call and then a **literal `2`** for the
`getExcessGibbsEnergy` call, on the very next line:

```java
public double getActivityCoefficientInfDil(int k) {
  PhaseInterface dilphase = this.clone();
  dilphase.addMoles(k, -(1.0 - 1e-10) * dilphase.getComponent(k).getNumberOfMolesInPhase());
  dilphase.getComponent(k).setx(1e-10);
  dilphase.init(dilphase.getNumberOfMolesInPhase(), dilphase.getNumberOfComponents(), 1, dilphase.getType(), 1.0);
  ((PhaseGEInterface) dilphase).getExcessGibbsEnergy(dilphase, 2, dilphase.getTemperature(), dilphase.getPressure(),
      dilphase.getType());
  return ((ComponentGEInterface) dilphase.getComponent(0)).getGamma();
}
```

`init` is handed `dilphase.getNumberOfComponents()`. `getExcessGibbsEnergy` is handed `2`.
**Every `PhaseGE` subclass loops on that argument** - `PhaseGENRTL`, `PhaseGEWilson`,
`PhaseGEUniquac`, `PhaseGEUnifac`, `PhaseDesmukhMather` and `PhaseDuanSun` all read
`for (int i = 0; i < numberOfComponents; i++) { ... phase.getComponent(i) ... }` - so a
phase with one component is asked for `getComponent(1)` and the answer is `null`, or one
past the end of the array.

Three subclasses crash three different ways, all from that one line:

| system | exception |
|---|---|
| `SystemNRTL` with CO2 alone | `ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1` at `ComponentGeNRTL.getGamma` |
| `SystemGEWilson` with CO2 alone | `NullPointerException` at `ComponentInterface.getx()` |
| `SystemDuanSun` | `NullPointerException` at `ComponentInterface.isIsIon()` |

**The neighbour three lines up has it right.** `getActivityCoefficientInfDilWater` passes
`2` on line 186 as well, and there `2` is correct, because the phase it evaluates was built
by `initRefPhases` with exactly two components (`addComponent(..., 0)` and
`addComponent(name, ..., 1)`). `getActivityCoefficientInfDil` clones **the phase itself**,
whose count is whatever the caller built. The literal looks right where it was copied from
and is wrong where it landed.

---

## How to reproduce

Self-contained; no test data needed. Three systems, one component each, all of them
component types whose `REFERENCESTATETYPE` is `solute` - which is what sends
`ComponentGE.fugcoef` down the infinite-dilution branch.

```java

public class OneComponentGe {
  private static void attempt(String label, neqsim.thermo.system.SystemInterface s) {
    s.addComponent("CO2", 1.0);
    s.setMixingRule("classic");
    try {
      s.init(0);
      s.init(1);
      System.out.printf("%-16s ok   gamma = %s%n", label, s.getPhase(1).getActivityCoefficient(0, 0));
    } catch (Throwable e) {
      System.out.printf("%-16s %s: %s%n", label, e.getClass().getSimpleName(), e.getMessage());
    }
  }

    attempt("SystemNRTL", new neqsim.thermo.system.SystemNRTL(313.15, 1.0));
    attempt("SystemGEWilson", new neqsim.thermo.system.SystemGEWilson(313.15, 1.0));
    attempt("SystemDuanSun", new neqsim.thermo.system.SystemDuanSun(313.15, 1.0));
  }
}
```

Output as it stands:

```
SystemNRTL       ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
SystemGEWilson   NullPointerException: ... getComponent(int) is null
SystemDuanSun    NullPointerException: ... getComponent(int) is null
```

`SystemDuanSun` is the one that cannot be avoided - its constructor adds a mole of `CO2`
and its `addComponent` refuses every other name, so **every** state it can hold is
one-component and water-free. `SystemNRTL` and `SystemGEWilson` need the caller to build a
one-component mixture deliberately, which is what the reproduction above does.

---

## Expected

A one-component phase evaluates at its infinite-dilution limit like any other: `2` is
replaced by the phase's own count, and the subclass sees the components that exist.

## Actual

The subclass is told there are two components, reads the second, and gets `null` or an
out-of-bounds index.

---

## Fix

```java
((PhaseGEInterface) dilphase).getExcessGibbsEnergy(dilphase, dilphase.getNumberOfComponents(),
    dilphase.getTemperature(), dilphase.getPressure(), dilphase.getType());
```

One expression, already written correctly on the line above. Line 186 does **not** need the
same change: that phase is built with two components and the literal matches.

---

## What the fix does not cover

`SystemDuanSun` remains unusable after this change, for a reason of its own: its
`addComponent` refuses water and the ions the model's correlation needs, so its salinity is
always zero. That is a separate report.

EvenSol:
Closes #3837.

The cloned phase used for infinite dilution was passed a component count of two regardless of its actual topology. Unary NRTL/Wilson/Duan-Sun evaluation could read an absent component; larger mixtures omitted components. The same method returned component zero even when the requested solute was another component.

Use the clone's actual count, return the requested solute's gamma, and normalize the trial mole fractions. Normalization is necessary for the unary Wilson limit: replacing the count alone otherwise gives gamma about 1e10 instead of 1.

Validation:
- Original master fails both new regressions: unary NRTL throws ArrayIndexOutOfBoundsException; a ternary NRTL case returns 1 instead of exp(0.5).
- Unary NRTL, Wilson and Duan-Sun now initialize and give gamma=1.
- A ternary NRTL case with alpha=0 and a single nonzero interaction has the independent analytical dilute limit ln(gamma_ethanol)=1/2. It checks the third solvent, nonzero solute index and unchanged source composition.
- Existing GE/hybrid and model-spec tests pass. Together with the independent model-conversion suite, 114 focused tests complete with one pre-existing skip.
- Repeated independently against original master with only this PR's production change; Java 8 compilation, repeated Spotless apply/check and documentation source audit pass.

AI-Assisted. Independent base a7ad601b2478757130d7b5f2f492337effc130a4. The separate unsupported Duan-Sun conversion problem is handled in another PR; this PR does not depend on it.

### CI repairs (2026-09-20)

- The three-compressor cooling example previously used binary feasibility search over a range that can be infeasible at both low and high rates because of surge/minimum-speed and maximum-capacity constraints. It now samples an explicit 0.5 percent flow grid in descending order and accepts only a freshly solved, feasible point within compressor limits. The positive production-gain assertions remain enabled.
- The Java 8 SVG process fixture failed a PS flash while compressing cooled, saturated gas. Suction scrubbers now remove condensed water/hydrocarbons before both stages. New assertions check single-phase compressor feeds, specified discharge pressures, positive finite powers, and the scrubbers' presence in the exported diagram. The original failure did not reproduce in the available Java 17 runtime; Java 8 confirmation remains pending.

Validation on OpenJDK 17:
- Focused cooling/SVG/GE tests: 4 passed.
- Complete cooling-analysis class plus SVG and GE regression classes: 17 passed, zero failures/errors/skips.
- Direct Spotless apply twice, Spotless check, documentation search audit, pre-commit and pre-push hooks, and whitespace checks passed.

Documentation impact: documented the non-monotonic compressor envelope and explicit sampling resolution in the bottleneck guide. Complete CI and Java 8 validation are **VALIDATION PENDING** on the updated head.


EvenSol:
CI investigation, 2026-09-20, exact head `37021717c5881dd59252666f6f92559c85768779`:

Slow partition 4 failed `CoolingDutyProductionAnalysisTest.test2026ScenarioCoolingAnalysisThreeCompressors` at the existing `Production should increase with cooling` assertion (job 106076352720). Its fixture constructs `SystemPrEos` with classic mixing, not the changed GE infinite-dilution path. No evidence currently attributes this failure to the GE change.

Controlled local checks in a separate worktree, with slow tests explicitly enabled using the repository's `-DexcludedTestGroups=` property:
- Original base `a7ad601b2478757130d7b5f2f492337effc130a4`: all 14 CoolingDutyProductionAnalysisTest tests pass, zero skips.
- Apply only this PR's exact two files, verifying their blob hashes: all 14 cooling tests and both GE regressions pass, zero skips.
- JDK 17 locally, changed code compiled with `--release 8`. Repeated direct Spotless apply/check and the documentation source audit pass; source hashes remain identical to this PR.

The failed CI job used Java 21. The failure remains unresolved in CI; these local passes do not prove it fixed or establish its environment/order sensitivity. Current-head Java 21 regular tests on Ubuntu and Windows passed, and Java 8 jobs are still running. GitHub does not permit the targeted rerun while the containing workflow remains in progress. Next dependency: rerun this failed job after workflow completion, then investigate any persistent failure with the same full-shard/JDK21 context. No production or test assertions were changed, and no branch update was published.

Documentation impact: none for this diagnostic-only follow-up.

EvenSol:
Post-merge verification recorded in #3837: https://github.com/equinor/neqsim/issues/3837#issuecomment-5751672114

The earlier cooling/SVG CI blockers are superseded: repaired head `d1c2565a1ae15e4f41f9743bf096af4a0a980bc8` passed 29 applicable checks, including all 14 cooling tests on Java 21 and both GE regressions plus the SVG fixture on Java 8 Windows. Those four Java files are unchanged in the merge and current master. Final head `f8484cac2aa44cbd88098ec16f66eaa5c1a0fd2f` also incorporated master; its full integration CI is still pending, so no all-green claim is made for that newer tree. No code or assertions were changed during this closure-verification pass.

PR #3841, opened 2026-09-20
[AI-Assisted] Reject unsupported Duan-Sun conversion and propagate model failures

### LINKED ISSUE #3839: `SystemDuanSun` holds one mole of CO2 whatever fluid it is given, so the Duan-Sun correlation never contributes

`neqsim/thermo/system/SystemDuanSun.java` lines 65 and 83-88, `SystemThermo.setModel` at
line 5365, `PhaseDuanSun.getExcessGibbsEnergy` at line 84,
`ComponentGeDuanSun.getGammaPitzer` at line 435.

---

## Summary

`SystemDuanSun`'s constructor adds a component by itself:

```java
addComponent("CO2", 1.0);          // line 65, unconditional
```

and its `addComponent` override refuses everything else:

```java
if (!"CO2".equals(componentName)) {
  throw new RuntimeException("SystemDuanSun supports only CO2");
}
```

So a `SystemDuanSun` holds **exactly one mole of CO2**, always, and cannot be given water,
an ion, or two of the three gases its own correlation carries.

That would be a defensible design if the model were a pure-CO2 equation of state. It is
not: `ComponentGeDuanSun.getGammaPitzer` is a **gas-in-brine Setschenow correlation**,

```
gamma = exp(2 S lambda_gas-Na + S^2 zeta_gas-NaCl)
```

with `lambda` and `zeta` fitted for CO2, nitrogen and oxygen, and `S` the salinity
`sum m_i` over the ions. Every input that expression needs is a component the system
refuses.

**Measured**: with the phase given one component, `getExcessGibbsEnergy` returns exactly
`0.0` and the activity coefficient is exactly `1.0`. There are no ions, so `S` is zero and
the correlation contributes nothing. The model is ideal for every state it can hold.

---

## How to reproduce

Self-contained; no test data needed.

```java
import neqsim.thermo.phase.PhaseDuanSun;
import neqsim.thermo.system.SystemSrkEos;

public class DuanSunHolds {
    SystemInterface s = new SystemSrkEos(313.15, 5.0);
    s.addComponent("water", 0.9);
    s.addComponent("Na+", 0.05);
    s.addComponent("Cl-", 0.05);
    s.setMixingRule("classic");
    s.init(0);
    System.out.print("the input fluid holds:");
    for (int i = 0; i < s.getNumberOfComponents(); i++) {
      System.out.print(" " + s.getComponent(i).getComponentName());
    }
    System.out.println();

    SystemInterface converted = s.setModel("Duan-Sun");
    System.out.print("setModel(\"Duan-Sun\") returns a " + converted.getClass().getSimpleName()
        + " holding:");
    for (int i = 0; i < converted.getNumberOfComponents(); i++) {
      System.out.print(" " + converted.getComponent(i).getComponentName() + " "
          + converted.getComponent(i).getNumberOfMolesInPhase() + " mol");
    }
    System.out.println();

    for (String name : new String[] {"water", "Na+", "nitrogen", "oxygen", "CO2"}) {
      try {
        new neqsim.thermo.system.SystemDuanSun(313.15, 5.0).addComponent(name, 0.1);
        System.out.println("  addComponent(\"" + name + "\") accepted");
      } catch (RuntimeException e) {
        System.out.println("  addComponent(\"" + name + "\") throws: " + e.getMessage());
      }
    }

    // With the component count the correlation would need, the phase is still ideal.
    PhaseDuanSun phase = new PhaseDuanSun();
    phase.addComponent("CO2", 1.0, 1.0, 0);
    phase.setTemperature(313.15);
    phase.setPressure(5.0);
    System.out.printf("  excess Gibbs energy = %.15g%n",
        phase.getExcessGibbsEnergy(phase, 1, 313.15, 5.0, phase.getType()));
    System.out.printf("  ln gamma(CO2)       = %.15g%n",
        ((neqsim.thermo.component.ComponentGeDuanSun) phase.getComponent(0))
            .getGammaPitzer(phase, 1, 313.15, 5.0, phase.getType(), 0.0) - 1.0);
  }
}
```

Output as it stands:

```
the input fluid holds: water Na+ Cl-
setModel("Duan-Sun") returns a SystemDuanSun holding: CO2 1.0 mol
  addComponent("water") throws: SystemDuanSun supports only CO2
  addComponent("Na+") throws: SystemDuanSun supports only CO2
  addComponent("nitrogen") throws: SystemDuanSun supports only CO2
  addComponent("oxygen") throws: SystemDuanSun supports only CO2
  addComponent("CO2") accepted
  excess Gibbs energy = 0.00000000000000
  ln gamma(CO2)       = 0.00000000000000
```

---

## A second thing in the same path

`SystemThermo.setModel` copies the source fluid's components into the new system one at a
time, inside a `try` whose `catch` logs and **returns the half-built system anyway**:

```java
} catch (Exception ex) {
  logger.error(ex.getMessage(), ex);
}
return tempModel;
```

So the `"SystemDuanSun supports only CO2"` exception, which exists to tell a caller exactly
what went wrong, never reaches one. `setModel("Duan-Sun")` on a brine returns a
`SystemDuanSun` holding one mole of CO2 and no indication that the composition was
discarded. The caller finds out at `init`, as a `NullPointerException` from a component
that does not exist - a symptom of the count literal in `PhaseGE` reported separately.

---

## Expected

Either the model computes what its correlation describes - the solubility of CO2, nitrogen
and oxygen in a brine - in which case `SystemDuanSun` must accept water, the ions and all
three gases; or it does not, in which case `"Duan-Sun"` should not be a name `setModel`
accepts.

## Actual

Every state it can hold is one mole of CO2 with an activity coefficient of exactly one.

---

## What would settle it

**Is `"Duan-Sun"` meant to be reachable from `setModel` at all?** It is the only name in
that dispatch whose system refuses the components it is handed - the other restrictive
systems (`SystemSpanWagnerEos`, `SystemWaterIF97`, `SystemAmmoniaEos`, `SystemLeachmanEos`)
have the same single-component guard but are not dispatch targets. Removing it from the
dispatch is one line and would leave the correlation reachable only through
`PhaseDuanSun` directly, which is how the class's own test uses it.

**If it is meant to be reachable**, the correlation needs `SystemDuanSun.addComponent` to
accept water, the ions and the gases rather than a hardcoded mole of CO2 - that is a
substantially larger change, and it is the one that makes `getGammaPitzer` do anything.

Either way the swallowed exception is worth fixing: a guard whose message cannot reach the
caller is not a guard.

EvenSol:
Closes #3839 using the report's explicit unsupported-conversion resolution.

setModel("Duan-Sun") constructed a prefilled one-mole CO2 system, rejected the first brine component, swallowed the exception and returned the incomplete system. This silently replaced the source composition.

Reject Duan-Sun system conversion with an explicit reason. Propagate all setModel conversion failures as IllegalArgumentException with the original cause, and report unknown model names directly. Supported conversions still copy the source inventory and state. Update the API contract, model/creation guides and reference manual to explain the limitation.

The historical SystemDuanSun constructor and direct PhaseDuanSun correlation API remain available for compatibility. This deliberately does not claim a validated brine model or enable unqualified salinity/species behavior.

Validation:
- Original master fails the new unsupported-conversion and invalid-name tests because it returns instead of throwing.
- Tests cover rejection without changing the source, null/unknown names, a component-copy failure after preceding components have been copied, and a successful PR conversion preserving every component amount, T and P.
- Model-conversion/SystemThermo tests pass independently with all other proposed production fixes restored to original master.
- Combined GE and conversion validation: 114 tests, one pre-existing skip; Java 8 compilation, repeated Spotless apply/check and documentation source audit pass.
- The shared baseline documentation version mismatch is repaired separately in #3836; full CI remains the integration gate.

AI-Assisted. Independent base a7ad601b2478757130d7b5f2f492337effc130a4. Does not depend on the GE infinite-dilution fix.


EvenSol:
Fixed the documentation-search failure in 535158f by applying the five-page current-release correction from #3836. All 385 documentation contract tests, direct Spotless apply/check, search source audit and both hook stages pass. No numerical changes; fresh CI pending.

PR #3910, opened 2026-09-22
Fix agent_search.py to discover CLI-installed agents by default

geonorway:
## Summary

`devtools/agent_search.py` — the semantic agent-discovery tool that
`router.agent.md` (the `@neqsim help` agent) documents as its recommended
fallback when a request doesn't match the hand-maintained routing table —
only indexed community/enterprise agents from **sibling git clones**
literally named `neqsim-community-agents` / `neqsim-enterprise-agents`.

It had no awareness of `~/.neqsim/agents`, the directory `neqsim agent
install` (the normal, documented way most users obtain community/private
agents) actually populates. As a result, an agent that is fully installed
and already invocable (e.g. via `@olga-simulation-agent`) was completely
invisible to this search, which could surface an unrelated false-positive
keyword match instead.

### How this was found

While running an OLGA oil-water pressure-drop case through the newly
installed `olga-simulation-agent` (installed via `neqsim agent install
olga-simulation-agent --vscode`, no local sibling clone of
`neqsim-community-agents`), I asked `@neqsim help "create and run an OLGA
case"` to sanity-check router discovery. Its documented discovery command:

```
devtools/agent_search.py "create and run an OLGA case" --top 8
```

returned `@safety-depressuring` as the top hit (a false-positive keyword
collision on "case") and never surfaced `olga-simulation-agent` at all,
even though it had been successfully invoked earlier in the same session.

## Fix

- Added `~/.neqsim/agents` as a **permanent default discovery root** in
  `_discover_roots`, mirroring `install_agent.py`'s `INSTALL_DIR` constant
  exactly (`Path.home() / ".neqsim" / "agents"`).
- Added a `NEQSIM_AGENTS_HOME` env-var override via a new
  `_installed_agents_root()` helper, so callers/tests can redirect this
  without depending on the real machine's `Path.home()`.
- Fixed a latent test-isolation bug this exposed: `CrossRepoDedupTest`'s
  synthetic `asset-economics-agent` fixture silently collided with a
  *real* agent of the same name already installed on the dev machine once
  the new default root was added — now redirected via `NEQSIM_AGENTS_HOME`.
- Added `InstalledAgentsRootTest` regression tests locking in the fix.
- Logged the finding in `docs/development/TASK_LOG.md` per the repo's
  continuous-improvement policy.

## Verification

```
$ python devtools/agent_search.py "create and run an OLGA case" --top 3
Top 3 agents for: 'create and run an OLGA case'

  1.000  olga-simulation-agent   (@olga-simulation-agent) [installed]  loads: neqsim-olga-multiphase-simulator
  0.473  run neqsim safety and depressuring simulation (@safety-depressuring) [neqsim]  ...
  0.325  cfd-coupling-agent      (@cfd-coupling-agent) [installed]  loads: neqsim-cfd-coupling
```

No flags needed — previously this query did not surface the agent at all.

```
$ python -m pytest devtools/test_agent_search.py -v
...
12 passed in 0.58s
```
(10 existing tests + 2 new regression tests, all green.)

## Files changed

- `devtools/agent_search.py` — new default discovery root + env override
- `devtools/test_agent_search.py` — hermetic fix + new regression tests
- `docs/development/TASK_LOG.md` — task log entry

No behavior change for users who already had the sibling repos cloned;
this purely adds a previously-missing discovery source.


EvenSol (commented):
## Review — does this improve NeqSim and the agents?

**Yes, the diagnosis is right and the direction is correct**, so I've kept it and pushed two fixes on top (commit `9abb76f`) rather than blocking. Verification below is measured on a real machine, not argued.

### What I confirmed

The root cause is real: `_discover_roots` only knew about sibling clones, so a CLI-installed agent was invisible. Tests pass, and `_load_from_dir` already no-ops on a missing directory, so the new default root is safe for everyone.

Catching the `CrossRepoDedupTest` collision with the *genuine* installed `asset-economics-agent` is a good find — that test's outcome depended on the developer's machine.

### What I found and fixed

**1. It missed the install path most users now have.** A marketplace plugin install keeps its agents inside the plugin package and leaves `~/.neqsim/agents` empty, so the motivating query still returned nothing there. Measured here:

```
~/.vscode/agent-plugins/.../neqsim{,-community,-enterprise}/com.github.copilot/agents   36 / 52 / 76
~/.neqsim/agents                                                                          0
```

Added `~/.vscode/agent-plugins` and `~/.copilot/installed-plugins` (flat `*.agent.md`), labelled from each `plugin.json` name. On a plugin-only machine the PR's own example now works with no flags:

```
1.000  olga-simulation-agent     [plugin:neqsim-community]
0.421  enterprise-thelma-agent   [plugin:neqsim-enterprise]
```

A sub-trap: `%APPDATA%/Code/agentPlugins` holds version-hash **cache copies** of the same plugins (`…/neqsim-community/1a0c50d1a81/`). Keyed on the folder name they looked like separate plugins and re-listed all 164 agents under labels like `plugin:1a0c50d1a81`. Resolving the name from `plugin.json` collapses them.

**2. Duplicate records when both sources exist.** Dedup is `(repo, name)`, and the installed copy carries a different label — so the same agent is indexed twice and burns two `--top N` slots. Reproduced by populating `~/.neqsim/agents` from the community clone:

```
total: 169  {enterprise: 76, community: 52, neqsim: 36, installed: 5}
names indexed twice: artificial-lift-agent, asset-economics-agent, cfd-coupling-agent,
                     community-router-agent, compressor-antisurge-agent
```

That is exactly the maintainer setup this repo assumes (sibling clones **and** `neqsim agent install --all --vscode --force`), so "no behavior change for users who already had the sibling repos cloned" doesn't hold. Installed/plugin roots are now *fallbacks*: a copy is skipped when a checkout already supplied that handle, while the `(repo, name)` rule still applies among fallbacks so the deliberate community/enterprise pair survives on a plugin-only machine. After the fix: 164 records with clones, 164 plugin-only, zero duplicates either way.

**3. `NEQSIM_AGENTS_HOME` was search-only.** `install_agent.INSTALL_DIR` was a plain constant, so the override moved where search *looks* but not where install *writes* — a docstring saying "mirrors exactly" is the classic setup for drift. `INSTALL_DIR` now reads the same variable.

Also added `NEQSIM_AGENT_PLUGINS_HOME` so the tests stay hermetic against the developer's own plugin tree (same failure mode as the `asset-economics-agent` collision, one level up), 5 regression tests (17 total, all green), and a line in `router.agent.md` naming the sources the retriever covers — the PR notes that gap but didn't close it.

### Left for you

- The diff also appends an unrelated **2026-09-21 "3-stage compression with intercooling"** entry at the bottom of `TASK_LOG.md` (not on master, and the file's rule is *"Add new entries at the top. Most recent first."*). Please split it out.
- The new TASK_LOG `Notes:` is one ~1,500-character paragraph that GitHub truncates in the diff; the PR body already carries the narrative. Worth trimming, and it should now mention the plugin root as well.
- Nit: `@return` is a Javadoc tag; no other docstring in `agent_search.py` uses it (I dropped it in the moved copy).

EvenSol (commented):
## Review — does this improve NeqSim and the agents?

**Yes, the diagnosis is right and the direction is correct**, so I've kept it and pushed two fixes on top (commit `9abb76f`) rather than blocking. Verification below is measured on a real machine, not argued.

### What I confirmed

The root cause is real: `_discover_roots` only knew about sibling clones, so a CLI-installed agent was invisible. Tests pass, and `_load_from_dir` already no-ops on a missing directory, so the new default root is safe for everyone.

Catching the `CrossRepoDedupTest` collision with the *genuine* installed `asset-economics-agent` is a good find — that test's outcome depended on the developer's machine.

### What I found and fixed

**1. It missed the install path most users now have.** A marketplace plugin install keeps its agents inside the plugin package and leaves `~/.neqsim/agents` empty, so the motivating query still returned nothing there. Measured here:

```
~/.vscode/agent-plugins/.../neqsim{,-community,-enterprise}/com.github.copilot/agents   36 / 52 / 76
~/.neqsim/agents                                                                          0
```

Added `~/.vscode/agent-plugins` and `~/.copilot/installed-plugins` (flat `*.agent.md`), labelled from each `plugin.json` name. On a plugin-only machine the PR's own example now works with no flags:

```
1.000  olga-simulation-agent     [plugin:neqsim-community]
0.421  enterprise-thelma-agent   [plugin:neqsim-enterprise]
```

A sub-trap: `%APPDATA%/Code/agentPlugins` holds version-hash **cache copies** of the same plugins (`…/neqsim-community/1a0c50d1a81/`). Keyed on the folder name they looked like separate plugins and re-listed all 164 agents under labels like `plugin:1a0c50d1a81`. Resolving the name from `plugin.json` collapses them.

**2. Duplicate records when both sources exist.** Dedup is `(repo, name)`, and the installed copy carries a different label — so the same agent is indexed twice and burns two `--top N` slots. Reproduced by populating `~/.neqsim/agents` from the community clone:

```
total: 169  {enterprise: 76, community: 52, neqsim: 36, installed: 5}
names indexed twice: artificial-lift-agent, asset-economics-agent, cfd-coupling-agent,
                     community-router-agent, compressor-antisurge-agent
```

That is exactly the maintainer setup this repo assumes (sibling clones **and** `neqsim agent install --all --vscode --force`), so "no behavior change for users who already had the sibling repos cloned" doesn't hold. Installed/plugin roots are now *fallbacks*: a copy is skipped when a checkout already supplied that handle, while the `(repo, name)` rule still applies among fallbacks so the deliberate community/enterprise pair survives on a plugin-only machine. After the fix: 164 records with clones, 164 plugin-only, zero duplicates either way.

**3. `NEQSIM_AGENTS_HOME` was search-only.** `install_agent.INSTALL_DIR` was a plain constant, so the override moved where search *looks* but not where install *writes* — a docstring saying "mirrors exactly" is the classic setup for drift. `INSTALL_DIR` now reads the same variable.

Also added `NEQSIM_AGENT_PLUGINS_HOME` so the tests stay hermetic against the developer's own plugin tree (same failure mode as the `asset-economics-agent` collision, one level up), 5 regression tests (17 total, all green), and a line in `router.agent.md` naming the sources the retriever covers — the PR notes that gap but didn't close it.

### Left for you

- The diff also appends an unrelated **2026-09-21 "3-stage compression with intercooling"** entry at the bottom of `TASK_LOG.md` (not on master, and the file's rule is *"Add new entries at the top. Most recent first."*). Please split it out.
- The new TASK_LOG `Notes:` is one ~1,500-character paragraph that GitHub truncates in the diff; the PR body already carries the narrative. Worth trimming, and it should now mention the plugin root as well.
- Nit: `@return` is a Javadoc tag; no other docstring in `agent_search.py` uses it (I dropped it in the moved copy).

